PR #18, opened 2019-02-25
css now returns a className instead of a function

michael-klein:
Implemented the change for CSS we discussed. styled itself now knows how to handle the "no tag" situation. File size is as follows:

image

### DIFF
--- a/src/styled.js
+++ b/src/styled.js
@@ -33,5 +33,5 @@ export const styled = tag =>
         })
       );
     };
-    return !tag ? processStyles() : processStyles;
+    return tag ? processStyles : processStyles();
   };
--- a/src/styled.js
+++ b/src/styled.js
@@ -5,7 +5,7 @@ let h;
 
 /**
  * Sets custom pragma to be used in contexts
- * @param {function} val 
+ * @param {function} val
  */
 export const setPragma = val => (h = val);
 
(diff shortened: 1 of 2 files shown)


src/styled.js
```diff
@@ -14,7 +14,8 @@ export const setPragma = val => (h = val);
  * @param {String} tag DOM tagName
  * @return {Function}
  */
-export const styled = tag => (str, ...defs) => props => {
+export const styled = tag => (str, ...defs) => {
```
cristianbote: Awesome stuff @michael-klein!

I think we can shave some more bytes! 🎉 Spread operator is really compiled down to a `for` loop, and I just did a quick test, and we can get down to `919B`!

```js
export const styled = tag => function() {
  const args = Array.from(arguments);  
  const processStyles = props => {
    const className = getClassNameForCss(getCss(args.shift(), args, props));
```

src/styled.js
```diff
@@ -23,4 +24,6 @@ export const styled = tag => (str, ...defs) => props => {
     return h(tag, Object.assign({}, props, {
       className: (props && props.className ? props.className + " " : "") + className
     }));
+  }
+  return !tag ? processStyles() : processStyles
```
cristianbote: If you flip the `if` that's gonna remove the `!` operator and shave 1B 😄

src/styled.js
cristianbote: Actually, just tried it and it breaks the `args` values with `.shift()`. But using, `[0]` and a `.slice(1)` does the trick.

```js
export const styled = tag => function() {
  const args = Array.from(arguments);  
  const processStyles = props => {
    const className = getClassNameForCss(getCss(args.shift(), args, props));
```

PR #21, opened 2019-02-26
Use bind

michael-klein:
Implemented the ability to .bind pragma and target on styled/css.


src/core/style/sheet.js
```diff
@@ -1,52 +1,33 @@
...
+  styles += css;
+
+  // If we're not the client
+  if (target) {
+    let sheet = target.querySelector("style[" + SHEET_ID + "]");
+    if (!sheet) {
+      sheet = document.createElement("style");
+      sheet.setAttribute(SHEET_ID, "");
+      target.appendChild(sheet).innerHTML = css;
+    } else sheet.firstChild.data += css;
```
cristianbote: Just for readability can you add the `{` for else?

src/styled.js
```diff
@@ -15,20 +14,25 @@ export const setPragma = val => (h = val);
  * @param {String} tag DOM tagName
  * @return {Function}
  */
-export const styled = tag =>
-  function() {
+export const styled = function(tag) {
+  const styledContext = this || {};
+  const pragma = styledContext.pragma || h;
```
cristianbote: Maybe if we keep, for now, pragma from `h` that would shave some bytes as well.

src/core/style/sheet.js
```diff
@@ -1,52 +1,33 @@
...
- * @param {String} hash
  * @param {String} css
+ * @param {Element} target
  */
-export const add = (hash, css) => {
-
-  // If this is already present just stop
-  if (styles[hash] == css) {
+export function add(css, target) {
+  if (styles.indexOf(css) > -1) {
```
cristianbote: `styles.indexOf(css) > -1` -> `~styles.indexOf(css)` would save some bytes as well

src/core/style/sheet.js
```diff
@@ -1,52 +1,35 @@
...
  * Adds the hash and it's css to cache and to appends it to stylesheet
- * @param {String} hash
  * @param {String} css
+ * @param {Element} target
  */
-export const add = (hash, css) => {
-
-  // If this is already present just stop
-  if (styles[hash] == css) {
+export function add(css, target) {
```
cristianbote: Changing this back to `const` shaves 6B

PR #33, opened 2019-03-04
Adding TS types definition file.

pmkroeker:
This adds a `goober.d.ts` file to the `/src` directory. Also changed was adding `typings.json` to the files array, and adding reference the the project types in the `package.json`.

This is a fairly rudimentary pass at adding types. CSS properties are not checked for to ensure that they are passed proper values.

Also, my use case is the vanilla implementation, not using styled components. I believe they should still work, but overrides for `react` and `preact` implementations may be required.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "integrity": "sha512-OfC2uemaknXr87bdLUkWog7nYuliM9Ij5HUcajsVcMCpQrcLmtxRbVFTIqmcSkSeYRBFBRxs2FiUqFJDLdiebA==",
       "dev": true,
       "requires": {
-        "@babel/highlight": "7.0.0"
+        "@babel/highlight": "^7.0.0"
       }
     },
     "@babel/core": {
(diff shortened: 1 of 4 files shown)


pmkroeker:
Updated the PR. `csstype` has a whole hyphenated properties section. So now if you use object notation in TS you get some nice autocomplete features.

pmkroeker:
Also once this gets merged in I would be more than happy to provide a vanilla + typescript demo!

pmkroeker:
Also updated. `bundlesize` can't be install properly on my machine so I removed it so that the `package.json` and the `package-lock.json` files were in sync.

@cristianbote if you could possibly take this and re-add it in I'm sure all the tests would pass. Adding typings won't increase the size of the package.

pmkroeker:
I changed from included `bundlesize` to `size-check: npx bundlesize`.  Performs the same way on travis-ci, and allows me not to have to worry about the install not working and messing up the package on my machine.

pmkroeker:
I'm going to test the preact styled tag today. I think the typing might be off.

pmkroeker:
Nope no issues with using preact with setPragma and styled()!

cristianbote:
Awesome job @pmkroeker! Really helpful stuff. I was going to ask about the `npx` if it would be fine to change it back, but it's fine. Usually I don't use `npx` because of it's inner workings but I guess it's fine.

A HUGE thank you! 😄

pmkroeker:
If they need updating in the future I can keep the typings up to date. And ya you could change the bundlesize thing back if you want, for some reason it just doesn't install right for me.

Issue #35, opened 2019-03-07
Bug - Cannot find csstypes


pmkroeker:
As csstypes is listed as a devDependency it cannot be found when trying to import to the types file.

> node_modules/goober/src/goober.d.ts:1:57 - error TS2307: Cannot find module 'csstype'.
> 
> 1 import { PropertiesHyphen as CSSPropertiesHyphen } from 'csstype';

pmkroeker:
csstype has been moved to a peerDependency. Thanks @cristianbote!

PR #36, opened 2019-03-07
Moved csstype from devDependency to dependency

pmkroeker:
This should resolve issue #35. I moved `csstype` from being a dev-dependency to a regular dependency. The other solution is to have it as a peerDependency, but I think this solution is better. Still adds nothing to bundle size as its imported in `goober.d.ts`


pmkroeker:
Fixed with move to peer!

PR #41, opened 2019-03-08
bugfix/object hashing

### LINKED ISSUE #39: css returning same hash for different properties.
It seems like every call to css is outputting the same hash.

```typescript
const tickLabel = css({
    'font-size': '1em',
    'cursor': 'default',
    'user-select': 'none',
    '>.tick>line': {
        'stroke-width': '1px',
        'stroke-opacity': 0.5,
        'shape-rendering': 'crispEdges',
    },
    '>path': {
        'stroke-width': '2px',
    },
});

const gridStyle = css({
    '>path': {
        stroke: 'none',
    },
    '>.tick>line': {
        'stroke-width': '1px',
        'stroke-opacity': 0.5,
    },
});
console.log(tickLabel, gridStyle);
```
outputs `g015475 g015475`

Could this be related to the caching update? Seemed to be working properly before that was added.

pmkroeker:
This fixes #39 . I added a JSON.stringify to the compiled css string. There is an issue in production where if compiled is an Object the cache key is set to `[object Object]` and the next object that is used returns the cached className.

Adds a few bytes but preserves correct functionality using object notation instead of strings.

### DIFF
--- a/src/core/hash.js
+++ b/src/core/hash.js
@@ -18,7 +18,7 @@ let cache = {
 export const hash = (compiled, target, glob) => {
     // generate hash
     const compString = JSON.stringify(compiled);
-    const className = cache[comp_string] || (cache[comp_string] = glob ? "" : toHash(compiled));
+    const className = cache[compString] || (cache[compString] = glob ? "" : toHash(compiled));
 
     // Parse the compiled
     const parsed = cache[className] || (cache[className] = parse(
--- a/src/core/hash.js
+++ b/src/core/hash.js
@@ -17,7 +17,7 @@ let cache = {
  */
 export const hash = (compiled, target, glob) => {
     // generate hash
-    const comp_string = JSON.stringify(compiled);
+    const compString = JSON.stringify(compiled);
     const className = cache[comp_string] || (cache[comp_string] = glob ? "" : toHash(compiled));
 
     // Parse the compiled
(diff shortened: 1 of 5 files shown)


src/core/hash.js
```diff
@@ -17,7 +17,8 @@ let cache = {
  */
 export const hash = (compiled, target, glob) => {
     // generate hash
-    const className = cache[compiled] || (cache[compiled] = glob ? "" : toHash(compiled));
+    const comp_string = JSON.stringify(compiled);
```
cristianbote: Great finding! 👍 

Would you mind changing the `const` name to camelCase? 😄

cristianbote: Also, I think a test case for this would be helpful to not run into this issue, again.

pmkroeker: Oh ya I can do that. Been writing a bunch of python so used the wrong standard lol.

pmkroeker: Not exactly sure on how I should test it. I'll have a bit of time shortly to look into it and get a proper test. Maybe just it twice with 2 objects and ensure that the return is different?

pmkroeker: Fixed the casing and the references.

I updated the hash test at the end to do a check to ensure that returned vals are different for different objects. Let me know if it should change.

pmkroeker:
Just tested the files in my charting library and can confirm this fixes the issue!

PR #43, opened 2019-03-13
Extract css any target

michael-klein:
extractCSS should now work in any situation:

- in SSR environment without DOM: same as before
- in SSR environment with DOM (headless chrome etc.) same as above (didn't work at all before)
- with different targets than document.head: exractCSS now supports optional target argument and will extract from that target

Also, bytes saved:
1012B UMD (with current master merged in)


cristianbote:
Awesome job @michael-klein! Do you think we can add something inside README as well? To make it easier to grasp the concept?

PR #47, opened 2019-04-20
Nested with multiple selector

UpperCod:
This small change allows you to use nested with multiple selectors.

I have minimally affected the core, but I recommend:

that the parameter paren parse, is always an array, that will prevent the use of concat
```js
/**
* @param {string[]} paren 
**/
export const parse = (obj, paren, wrapper) 
```

### DIFF
--- a/src/core/parse.js
+++ b/src/core/parse.js
@@ -8,7 +8,7 @@ export const parse = (obj, paren, wrapper) => {
     let current = "";
     let blocks = "";
     let outer = "";
-    
+    paren = [].concat(paren);
     // If we're dealing with keyframes just flatten them
     if (/^@k/.test(wrapper)) {
       // Return the wrapper, which should be the @keyframes selector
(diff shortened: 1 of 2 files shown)


cristianbote:
Hi @UpperCod! Thanks for submitting this PR. I don't see an issue changing the type for `paren`. The only thing would be the size. So that's what I'm trying to look around and see if I can find stuff to move around.

We'll get back asap.

UpperCod:
I apologize github does not notify me anything, I will see how I help the code reduction, in my fork omiti the use of bundlesize, since my node version could not specify the installation with an error in node-gyp, if you like, share the node version that you have used to verify if it is a Node or OS version error in my installation

cristianbote:
Just a reminder, that I'm still looking into this one 😄 I was out for holidays and I'm back at it. Keep you posted!

src/core/parse.js
```diff
@@ -24,18 +24,20 @@ export const parse = (obj, paren, wrapper) => {
         if (isImport) {
             outer += key + " " + val + ";";
         } else if (typeof val === "object") {
-
-            // Regular selector
-            let next = paren + " " + key;
-            
-            // Nested
-            if (/&/g.test(key)) next = key.replace(/&/g, paren);
```
cristianbote: So, I had some time today and gave it a try. With the recent code shave, if you add a new condition just for this, that would keep the size under `1024B`.

```js
if (/,/.test(paren)) next = next.replace(/,/g, ` ${key},`);
```

What do you think?

UpperCod: It seems an interesting idea, but even so you will have to decompose the `paren` and the `key`, in order to successfully nest each selector.

paren: `parent-1, parent-2` key:` &: hover, &: active, span`

```css
parent-1,
parent-2{
   color:red;
  &:active,
  &:hover,
   span{
        color:black;
  }
}
```

I have not tried the code inside goober, but I think this would happen, the expression prevents nested from being applied to the last  paren(`parent-2`) selector and only nested key to the selector that it conserved with the expression, but `&: active, span` would be badly nested... I will see how to reduce the code that I have shared

cristianbote:
Hey @UpperCod it's been a while 😄I'm thinking of opening an issue to track this and commit the test that you've added. What do you think?

UpperCod:
> Hey @UpperCod it's been a while 😄I'm thinking of opening an issue to track this and commit the test that you've added. What do you think?

I think it's excellent, but I think the ideal is to eliminate the effect of  `&` for multiple selectors and replace it with `:is`, this would facilitate the implementation.

cristianbote:
Added the test in #140. Thank you @UpperCod for this one!

PR #50, opened 2019-10-01
Create a new object and assign props to it for a new extensible object

mjanssen:
When using the `styled(tag)` function, the props passed to the component are assigned to `_ctx.p` and new properties are assigned to this object. However, `_ctx.p` is unable to be extended, due to `preventExtensions`. This results in an error `cannot add property className, object is not extensible`.

Creating a new object and assigning the props to it, will result in an extensible object again.

### DIFF
--- a/src/styled.js
+++ b/src/styled.js
@@ -14,7 +14,7 @@ function styled(tag) {
     const _args = arguments;
 
     return function Styled(props) {
-      _ctx.p = props ? Object.assign({}, props) : {};
+      _ctx.p = Object.assign({}, props);
       _ctx.o = /\s*go[0-9]+/g.test(_ctx.p.className);
       _ctx.p.className = css.apply(_ctx, _args) + (_ctx.p.className ? " " + _ctx.p.className : "");
 
--- a/src/styled.js
+++ b/src/styled.js
@@ -10,11 +10,11 @@ const setPragma = pragma => (h = pragma);
 function styled(tag) {
   const _ctx = this || {};
 
-  return function() {
+  return function () {
     const _args = arguments;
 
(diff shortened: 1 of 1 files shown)


src/styled.js
```diff
@@ -10,11 +10,11 @@ const setPragma = pragma => (h = pragma);
 function styled(tag) {
   const _ctx = this || {};
 
-  return function() {
+  return function () {
     const _args = arguments;
 
     return function Styled(props) {
-      _ctx.p = props || {};
+      _ctx.p = props ? Object.assign({}, props) : {};
```
cristianbote: Should we just use the assign call directly? Maybe that'll shave some more bytes 🙂

mjanssen: I believe that should be possible. The props argument is always an object for me, so checking it is not required.

cristianbote:
Uh, and also the size is over 1KB 🙁 any way we could do this without the size increasing that much? 🤔

mjanssen:
I'm not entirely sure how we can maintain the size. The bundle increases with 8 bytes in total (1019 > 1028).

Which tests were you suggesting?

mjanssen:
Regarding the test; it's all tested with the tests already included (in __tests__/styled.test.js).

cristianbote:
Yeah, tricky one. Alright, let's get this in and I'll figure it out :) Thank you so much for this! Really appreciate your time for doing this

cristianbote:
Ah, just realised this is happening in `react` right? :) `preact` seems fine.

mjanssen:
That's correct. React seems to lock the props object, making it immutable after passing it further down.

PR #53, opened 2019-10-02
Allow passing in props to styled elements.

### LINKED ISSUE #52: Type definitions for styled() functions
When using typescript with styled component libraries, you can pass the type of the props to the styled Component, to enforce correct property names and having autocompletion for the props object.

**Example**
```
type ParagaphType = {
  style?: object;
  color?: string;
};

const P = styled('div')<ParagaphType>`
  color: ${props => (props.color ? props.color : 'black')};
  margin: 0;
  padding: 0;
  ${props => props.style && { ...props.style }}
`;
```

With Goober, I'm getting the following typescript problem when implementing the same syntax.

image

Could we update the type defs for Goober to have the autocompletion as well?

pmkroeker:
Fixes #52. Allows people to pass prop interfaces through to the end component.

### DIFF
--- a/goober.d.ts
+++ b/goober.d.ts
@@ -5,12 +5,12 @@ export = goober;
 export as namespace goober;
 
 declare namespace goober {
-    function styled <T = any>(tag: string): tagged<T>;
+    function styled <T = {}>(tag: string): tagged<T>;
     function setPragma (val: StyledVNode): void;
     function extractCss (): string;
     function css (tag: CSSAttribute | TemplateStringsArray): ClassName;
-    type StyledVNode = <T>(props: T, ...args: any[]) => any;
-    type tagged = <T>(tag: CSSAttribute | TemplateStringsArray) => StyledVNode<T>;
+    type StyledVNode<T = {}> = (props: T, ...args: any[]) => any;
+    type tagged<T = {}> = (tag: CSSAttribute | TemplateStringsArray) => StyledVNode<T>;
     interface CSSAttribute extends CSSProperties {
         [key: string]: CSSAttribute | string | number;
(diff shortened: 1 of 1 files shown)


goober.d.ts
```diff
@@ -5,12 +5,12 @@ export = goober;
 export as namespace goober;
 
 declare namespace goober {
-    function styled (tag: string): tagged;
+    function styled <T = any>(tag: string): tagged<T>;
```
cristianbote: Ha, tried to paste your changes and see how it all works together but no luck. 🤔 
Screenshot 2019-10-03 at 11 12 41

pmkroeker: Dang I was hoping it was that easy. I have some time this morning so I'll take a closer look.

pmkroeker: I had placed the generics improperly. Fixed!

cristianbote:
Hey @pmkroeker hats-off to you! 😄 I have no idea what to look for in here though. @mjanssen can you test out Peter's PR? Is this working how one would expect?

Thanks to both of you!

pmkroeker:
This should allow you to pass props in like
```typescript
styled<Props>('div')`styles...`
// OR
styled('div')<Props>`styles...`
```

cristianbote:
@pmkroeker is there a way we could setup a test for these types definitions? 🤔

pmkroeker:
Yep, I can add that to this pr when I get some time.

pmkroeker:
I ended up working on the master branch of my fork, not my patch one (oops). So I'll close this PR and open a new one with that branch. I'll be adding the tests shortly.

PR #55, opened 2019-10-09
Add passing in prop generics + TS tests

### LINKED ISSUE #52: Type definitions for styled() functions

pmkroeker:
This is a re-open of #53 which fixes #52 . This will also add some preliminary ts testing.


cristianbote:
Awesome stuff @pmkroeker!

ts-tests/StyledComp.tsx
```diff
@@ -0,0 +1,20 @@
...
+import { styled, setPragma } from '../goober';
+
+setPragma(h);
+
+interface IButtonProps {
+    clicked: boolean;
+}
+
+const Button = styled<IButtonProps>('button')`
+    background: blue,
```
cristianbote: This looks nice! 😃  yay! 🎉 

Question, @mjanssen issue is actually about not being able to use the props as the given type. In this case, that would be `IButtonProps`. Something like this:

```tsx
const Button = styled<IButtonProps>('button')`
    background: ${props => props.clicked ? 'blue' : 'white' };
`
```

I think basically could we have some sort of a test that asserts that using a property that's not on the interface throws?

pmkroeker: Done! Typings are always fun!
I had to put the generic specification on the styled function as it was easier.
```typescript
const Button = styled<IButtonProps>('button')`
    background: ${props => props.clicked ? 'blue' : 'white' };
`
```

This case has a ts test in now.

PR #60, opened 2019-12-02
Typo: package -> goober

rs-ds:


README.md
```diff
@@ -67,7 +67,7 @@ These are not yet measured. Need some time.
 ## SSR
 The benchmark is testing the folowing scenario:
 ```jsx
-import styled from 'package';
+import { styled } from 'goober';
```
cristianbote: Hey @rs-ds! Thank you for opening this PR! Really appreciate it!

Ha, that section in there, it references the benchmark test for each package, `@emotion/styled`, `styled-components` and `goober`. So it's basically just a placeholder.

Maybe change it into
```js
import styled from 'styled-components|@emotion/styled|goober';
```

What do you think?

rs-ds: Ah. Makes sense. Apologies on late response. Somehow missed the notification :stuck_out_tongue: will push change asap.

cristianbote: Now that I'm thinking about it, this would make introduce even more confusion. `goober` does not have a default export, which is the `styled` method, so it would be wrong to add it to the import name. Honestly, I'm inclining to say that we should leave it as it is.

What do you think? How do you feel about it?

cristianbote:
Hey @rs-ds closing this since it is stale and also I think could it could stay the same. How do you feel about it?

PR #64, opened 2019-12-10
docs: tweak/expand css docs

Hotell:


README.md
```diff
@@ -168,8 +168,47 @@ const BtnClassName = css`
...
+`;
+
+// vanilla JS
+// BtnClassName({size:20}) -> g016360
+const btn = document.querySelector("#btn");
+btn.classList.add(BtnClassName({size:20}));
+
+// JSX
+// BtnClassName({size:20}) -> g016360
+const App => <button className={BtnClassName({size:20})}>click</button>
```
cristianbote: I guess there's a set of braces missing?
```suggestion
const App = () => <button className={BtnClassName({size:20})}>click</button>
```

README.md
```diff
@@ -168,8 +168,47 @@ const BtnClassName = css`
...
+```
+
+**💡NOTE**
+
+> If you provide props as an object within `css`, make sure to provide empty object as a default parameter, otherwise your app will throw error if you won't provide an object argument.
```
cristianbote: Thank you for this! 😃

README.md
Hotell: aaah right! thx!

PR #66, opened 2019-12-10
chore: add prettier and format script task

Hotell:
As a contributor I found extremely difficull to write any code and then format it by hand...
This PR adds prettier, which is a common standard used across oss projects.

- if you're ok with this, I'll add husky and lint-staged so everything will be formatted on every commit.

Lemme know.

cheers

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,7 @@
+{
+  "trailingComma": "none",
+  "tabWidth": 4,
+  "semi": true,
+  "singleQuote": true,
+  "printWidth": 100
+}
--- a/package.json
+++ b/package.json
@@ -26,7 +26,8 @@
     "build": "npm run clean && microbundle --entry src/index.js --output dist --name goober --no-sourcemap && npm run size-check",
     "debug": "microbundle --entry src/index.js --output debug --name goober --no-sourcemap --no-compress",
     "dev": "npm run clean && microbundle watch --entry src/index.js --output dist --name goober",
-    "deploy": "npm run build && npm publish"
+    "deploy": "npm run build && npm publish",
+    "format": "prettier \"**/*.{js,ts,tsx,md}\" --write"
   },
   "keywords": [
     "css-in-js",
@@ -51,9 +52,12 @@
     "benchmark": "^2.1.4",
     "bundlesize": "^0.18.0",
     "csstype": "^2.6.3",
+    "husky": "3.1.0",
     "jest": "^24.1.0",
+    "lint-staged": "9.5.0",
     "microbundle": "^0.11.0",
     "preact": "^10.0.0",
+    "prettier": "1.19.1",
     "react": "^16.12.0",
     "react-dom": "^16.12.0",
     "styled-components": "^5.0.0-rc.2",
@@ -62,6 +66,7 @@
   "peerDependencies": {
     "csstype": "^2.6.2"
   },
+  "dependencies": {},
   "typings": "./goober.d.ts",
   "bundlesize": [
     {
@@ -69,5 +74,15 @@
       "maxSize": "1026B"
     }
   ],
-  "dependencies": {}
+  "lint-staged": {
+    "*.{js,ts,tsx,md}": [
+      "prettier --write",
+      "git add"
+    ]
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged"
+    }
+  }
 }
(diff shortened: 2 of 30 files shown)


cristianbote:
Yes! Wanted to do this for a while, but didn't had time for it. Thank you for taking the time to open this PR. I guess we should `husky` and the `lint-staged` hooks. That way we could bring all these together.

Hotell:
kk I'll push it in.

.prettierrc
```diff
@@ -0,0 +1,4 @@
+{
+  "semi": true,
+  "singleQuote": true
```
Hotell: **two question:**

- what style do you prefer ?
- should I run format on whole codebase, or you wanna keep it to be re-formatted gradually

package.json
```diff
@@ -26,7 +26,8 @@
     "build": "npm run clean && microbundle --entry src/index.js --output dist --name goober --no-sourcemap && npm run size-check",
     "debug": "microbundle --entry src/index.js --output debug --name goober --no-sourcemap --no-compress",
     "dev": "npm run clean && microbundle watch --entry src/index.js --output dist --name goober",
-    "deploy": "npm run build && npm publish"
+    "deploy": "npm run build && npm publish",
+    "format": "prettier \"**/*.{js,ts,tsx,md}\" --write"
```
cristianbote: Do you feel like we would still need this script, since `lint-staged` would take care of the formatting? I'm fine with it, just making sure this won't introduce another level of "wait, should I use format? Or..?"

.prettierrc
cristianbote: Thank you for asking! 😃I usually go with:
```json
{
    "trailingComma": "none",
    "tabWidth": 4,
    "semi": true,
    "singleQuote": true,
    "printWidth": 120
}
```

What do you think?

Hotell: I've quite opposite style :)
anyways these things are handled by robots, so I don't care :) 

going to apply your setup. thx!

package.json
Hotell: depends.

I usually keep this script for following scenarios:
- someone might skip the hook, no checking on CI, merge -> handy for reformat
- bumping to new versions of prettier that might introduce some regressions, run `format` done

Contributors should not care and follow pre-commit pipeline without noticing.

.prettierrc
Hotell: one suggestion though, I'd change "printWidth" to 100 or 80. 120 long lines is really long... :D

Hotell:
changes applied.

should I run format on whole codebase, or you wanna keep it to be re-formatted gradually

.prettierrc
cristianbote: :D I'm really open, tbh. As you said, it's _automated_, no need to worry about what goes where. So, yup, 100, 80, I have no preference. Your call here. 👍

package.json
cristianbote: Alright, fair enough. 👍

cristianbote:
Let's reformat everything 😄that way everyone could start on the same format. 👍

.prettierrc
```diff
@@ -0,0 +1,7 @@
+{
+  "trailingComma": "none",
+  "tabWidth": 4,
+  "semi": true,
+  "singleQuote": true,
+  "printWidth": 120
```
cristianbote: Don't forget about the `printWidth` 😉

Hotell: yup, done

Hotell:
allright, ready to re review and hopefully merge 🏖 thx

src/__tests__/styled.test.js
```diff
@@ -52,7 +52,7 @@ describe('styled', () => {
                 foo: 1;
             `(p)
         ).toEqual('h()');
-        expect(css).toBeCalledWith(['foo: 1']);
+        expect(css).toBeCalledWith([expect.stringContaining('foo: 1')]);
```
Hotell: this is fixing the tests, as prettier formats tagged `styled` literals to new lines, css function call won't get inline string

```diff
-styled("tag")`foo: 1`
+styled("tag")`
+  foo: 1
+`
```

3 solutions how to mitigate this:

1. using `expect.stringContaining` (this commit)
1. provide custom normalize function that removes trailing spaces + new lines
1. turn off prettier for that particular line and keep test as is

IMHO:  solution no 1, mirrors most of real life use cases of `styled` usage, thus is the best candidate for this test

Lemme know WDYT. thx

cristianbote: I think it's fine as you did it. We might change it, down the road but we'll get there. 👍 Solid work!

cristianbote:
@Hotell this is freaking awesome! The drive of the community always amazes me! 🙏

Thank you for this!

PR #67, opened 2019-12-10
use ===

cbbfcd:
thx the repo! 😄

### DIFF
--- a/src/core/compile.js
+++ b/src/core/compile.js
@@ -7,7 +7,7 @@ export const compile = (str, defs, data) => {
     return str.reduce((out, next, i) => {
         let tail = defs[i];
     
-        if (typeof defs[i] === "function") {
+        if (typeof defs[i] == "function") {
           const res = defs[i](data);
           const attr = res && (res.attributes || res.props);
           const end = (attr && attr.className) || (/^go/.test(res) && res);
--- a/src/core/compile.js
+++ b/src/core/compile.js
@@ -7,7 +7,7 @@ export const compile = (str, defs, data) => {
     return str.reduce((out, next, i) => {
         let tail = defs[i];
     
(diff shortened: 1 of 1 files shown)


src/core/compile.js
```diff
@@ -7,7 +7,7 @@ export const compile = (str, defs, data) => {
     return str.reduce((out, next, i) => {
         let tail = defs[i];
     
-        if (typeof defs[i] == "function") {
+        if (typeof defs[i] === "function") {
```
cristianbote: Hey @cbbfcd, thanks for taking the time to open this! Really appreciate it!

The use of double equal, it's meant to keep the size low while maintaining proper functionality. Have you encountered issues with the double and not the triple, type checking, as well?

Hotell:
time to add eslint for these kind of situations. so contributors will know it's ok :)

src/core/compile.js
cbbfcd: @cristianbote thx for your reply!

I fully understand the intention of using double equality, which is also safe in this scenario,
So [here](https://github.com/cristianbote/goober/blob/master/src/core/parse.js#L23) is a similar.

and it's can be replaced with `typeof val == 'object'`.

cbbfcd:
so we can add a filesize limit check, so every commit must be checked!

src/core/compile.js
cbbfcd: in fact, I prefer to use ===, the file size can be stolen from elsewhere

cristianbote: Hey @cbbfcd so, I've managed to shave `14B` and that meant to get rid of this exact check. There's the other one you've mentioned. If you'd like you can change that one instead.

cbbfcd: i will make a new PR about it. thx~

PR #70, opened 2019-12-11
docs: extend features list

Hotell:
this expands features list, with those that are not obvious from docs/examples 💪(but are present)


Hotell:
> Thank you 🙏 ! Hope you don't mind that I fixed the conflicts.

no worries, thx! I'll need to do another PR that reflects those features within typings as well

PR #63, opened 2019-12-10
Add support for forwardRef

donysukardi:
Fixes #22.

With this, user can setup `setForwardRef` like how they do `setPragma`, i.e.

```
setForwardRef(React.forwardRef)
```

~I'll add tests accordingly if this implementation is ok.~ Added tests

The few lines that I added seem to bring the bundle size to slightly over 1KB though. Any idea on how to bring the size down?
 
```
       1049 B: goober.js.gz
        938 B: goober.js.br
       1056 B: goober.module.js.gz
        944 B: goober.module.js.br
       1102 B: goober.umd.js.gz
        982 B: goober.umd.js.br

 FAIL  ./dist/goober.js: 1.02KB > maxSize 1KB (gzip)
 FAIL  ./dist/goober.module.js: 1.03KB > maxSize 1KB (gzip)
```

Issue #61, opened 2019-12-09
Theming solution?


nygdjs:
Thanks for this absolute awesomeness! I'm super excited, as on one of my projects we are currently try to find powerful, yet very lightweight solutions, and goober seems to fit in perfectly.
However I use styled-components on other projects for many reasons, but one of them surely is the powerful theming built-in. 
Do you plan something similar, or have any idea how to do something like that?

cristianbote:
Hey @nygdjs, thank you so much for your kind words! Really appreciate it!

So, theming is something that would increase the size of the bundle and would require `goober` to be specifically integrate with your framework. That means, `web-componens` could not make use of goober and `preact` for example would need to include the `preact/compat` layer to handle the theming part -- since that would rely on context api.

Do you see yourself in need of such api? Maybe I'm not seeing it, yet, what would theming help with. Could you give me an example?

Thank you so much for opening this!

donysukardi:
This seems doable if `gobber` provides a settable additional props getter/hook.

In gobber,

```jsx
let useAdditionalProps = () => ({})
const setUseAdditionalProps = fn => { useAdditionalProps = fn }

// in styled fn add the following
const additionalProps = useAdditionalProps();
const _props = Object.assign({}, props, additionalProps);
```

In user code,

```jsx
setUseAdditionalProps(() => {
  const theme = React.useContext(ThemeContext);
  return { theme }
})

// now props.theme is available to the styled components
```

Hotell:
> and preact for example would need to include the preact/compat layer to handle the theming part -- since that would rely on context api.

I had impression that context api is build in within Preact X no ? https://preactjs.com/guide/v10/context

Hotell:
I suggest to have theming provider as separate package to not increase lib size etc -> also it's always cleaner approach to have plugins 👌

cristianbote:
@Hotell yup, preact has the context api in core, but integrating the context api as imported from `React`, would mean other libs would have to add their compat layers to mock `React`.

donysukardi:
With my proposal above, the integration with context api is done on userland (via hook) so the library is still agnostic

cristianbote:
Just looked over, a second time. You are right 👍💯Do you wanna open a PR? 😄
Also, how about we merge the `setPragma` and `setTheme`? Hopefully we could save some more bytes

donysukardi:
What do you think of having a `setup({ pragma, forwardRef, useTheme })` function instead?

[...]

PR #76, opened 2019-12-17
Add setup function for pragma, forwardRef, and theme

donysukardi:
**Pending tasks**

- [ ] Tests
- [ ] Documentation
- [ ] Bundle size

This change increases the bundle size though. Any idea how we could trim it to fit the maxSize?

```
 FAIL  ./dist/goober.js: 1.03KB > maxSize 1KB (gzip)

 FAIL  ./dist/goober.module.js: 1.03KB > maxSize 1KB (gzip)
```

Related Issues: Theming (#61), forwardRef (#63)


src/styled.js
```diff
@@ -1,29 +1,37 @@
...
 function styled(tag) {
     const _ctx = this || {};
-
     return function() {
         const _args = arguments;
-
-        return function Styled(props) {
-            const _props = (_ctx.p = Object.assign({}, props));
+        function Styled(props, ref) {
+            const theme = useTheme();
```
cristianbote: Does it has to be a function? Could it be a plain object? Could become a perf bottleneck on user land.

src/styled.js
```diff
@@ -1,29 +1,37 @@
...
+            const _props = (_ctx.p = Object.assign(theme ? { theme } : {}, props));
             const _previousClassName = _props.className;
 
             _ctx.o = /\s*go[0-9]+/g.test(_previousClassName);
             _props.className =
                 css.apply(_ctx, _args) + (_previousClassName ? ' ' + _previousClassName : '');
+            if (ref) {
+                _props.ref = ref;
```
cristianbote: If we drop the `if` and always do `_props.ref = ref;` would that be strange? 🤔

src/styled.js
donysukardi: This is to get feature parity with other CSS-in-JS theming capability, i.e.
```jsx
const useTheme = () => useContext(ThemeContext)
```

Here's how it's similarly done in `styled-components@5` - https://github.com/styled-components/styled-components/blob/v5/packages/styled-components/src/models/StyledComponent.js#L129

donysukardi: For non context user, it could be just `() => theme`

src/styled.js
donysukardi: In terms of functionality, I don't think it'd be any different. As `ref` is treated as special props, it won't showed up in component props anyway.

cristianbote:
This is really great! Really love it! How do you feel about it? Can you add the tests and everything? What can I help you with?

src/styled.js
```diff
@@ -1,29 +1,37 @@
 import { css } from './css';
 
-let h;
-const setPragma = pragma => (h = pragma);
+let h, forwardRef, useTheme;
+const setup = (_h, _forwardRef, _useTheme = () => {}) => {
```
cristianbote: How about not defining a default `_useTheme`? Would that break?

src/styled.js
```diff
@@ -1,29 +1,37 @@
...
     const _ctx = this || {};
-
     return function() {
         const _args = arguments;
-
-        return function Styled(props) {
-            const _props = (_ctx.p = Object.assign({}, props));
+        function Styled(props, ref) {
+            const theme = useTheme();
+            const _props = (_ctx.p = Object.assign(theme ? { theme } : {}, props));
```
cristianbote: Also what do you think?
```suggestion
            const _props = (_ctx.p = Object.assign(useTheme ? useTheme() : {}, props));
```

cristianbote:
Did I mention how exciting this looks? 😄 Can't wait to have the theme support in. I can see myself using it, for sure!

This is a good one! Really appreciate your time on this! 👍

src/styled.js
donysukardi: Technically it's possible, but usually `useTheme` just returns `theme` instead of `{ theme }`, e.g. https://github.com/styled-components/styled-components/blob/v5/packages/styled-components/src/hooks/useTheme.js

It'd have to be like this instead,
```
const _props = (_ctx.p = Object.assign(useTheme ? { theme: useTheme() } : {}, props));
```

Let me give it a spin and see how the bundle size looks like, coz with gzip more code != bigger size and vice-versa :grin:

donysukardi:
Done the changes for both `useTheme` and `ref`. Bundle size seems to be lower now.

``` 
FAIL  ./dist/goober.js: 1.02KB > maxSize 1KB (gzip)

 FAIL  ./dist/goober.module.js: 1.01KB > maxSize 1KB (gzip)
```

I'll work on the tests for this.

nksaraf:
Hi,
I've been following the development of goober and absolutely love the simplicity of the code base. I have had to implement similar functionality for a fork. Is there anything I can do to help land this PR, or in other ways with the library.

cristianbote:
Hey @nksaraf thanks for checking goober all along! 🙂 Well anything that comes to mind! There a couple more things inside the issues that can be tackled as well. Give them a try see what you can do.

Thank you for your time!

donysukardi:
@cristianbote I've updated the implementation slightly. I observed that in styled-components, the `theme` from context is not passed down to the component, i.e. it's just used to generate the className with necessary styles.

I've also added unit and integration tests for the changes.

ryansolid:
This was going strong up to a month or so ago. Is this PR still moving forward? Is something holding it up? Are we still holding out hope it can get 0.02 kb smaller?

cristianbote:
Hey @ryansolid! :) Thanks for stopping by!

Well, it's just really my lack of time to do a perf bench on this and check the sizes. Also, this will definitely be a major version upgrade, since the whole `setPragma` update.

cristianbote:
Oh and documentation about the changes.

cristianbote:
Had a look into this change sets and it seems that this would add some considerable size. Something like `1035B`. Will dig more into it.

src/styled.js
```diff
@@ -1,29 +1,40 @@
...
-            _props.className =
-                css.apply(_ctx, _args) + (_previousClassName ? ' ' + _previousClassName : '');
+            const _props = {
+                ...props,
+                ref,
+                className:
+                    css.apply(_ctx, _args) + (_previousClassName ? ' ' + _previousClassName : '')
+            };
 
             return h(tag, _props);
```
JoviDeCroock:
```suggestion
            return h(tag, Object.assign({ ref, className: css.apply(_ctx, _args) + (_previousClassName ? ' ' + _previousClassName : '')  }, props));
```

This removes the need for line 26-31

src/styled.js
```diff
@@ -1,29 +1,40 @@
...
  * Styled function
  * @param {String} tag
  */
 function styled(tag) {
     const _ctx = this || {};
-
     return function() {
         const _args = arguments;
+        function Styled(props, ref) {
+            _ctx.p = {
```
JoviDeCroock:
```suggestion
            _ctx.p = Object.assign({ theme: useTheme ? useTheme() : undefined }, props)
```

cristianbote:
Hey @donysukardi things are happening! 🥳😎 Thanks to @JoviDeCroock for pushing this with his PR for autoprefixer. Overall this will end up under 1024B 🥳👍🥜 wohoo! So excited!

This will definitely be v2

cristianbote:
The tests failures are my doing 😑 sorry. I resolved the conflicts but somehow missed stuff. Will correct then asap

donysukardi:
This is awesome! I am excited!! Let me know if there's anything else I can help with :D

cristianbote:
Awesome! @donysukardi I think @JoviDeCroock opened a PR against yours to fix the test failures. In here https://github.com/donysukardi/goober/pull/1/files I think that would make this to pass.

cristianbote:
Alrighty! 😄 everything is fine. The above failures are expected 👍 

Thank you soo much @donysukardi! 🎉 🌟

PR #77, opened 2019-12-17
(golf) Drop the stringify from parse (-15B)

cristianbote:
Alrighty, more golfing! 🏌 

This PR drops the `JSON.stringify` for flattening the `@keyframes` and `@font-face` and use `parse` with special args. By doing that, the perf increases.
Screenshot 2019-12-17 at 16 28 04

Also, this is so awesome cause that would mean we could support more stuff.

Shaves `-15B`

### DIFF
--- a/src/core/parse.js
+++ b/src/core/parse.js
@@ -9,19 +9,6 @@ export const parse = (obj, paren, wrapper) => {
     let blocks = '';
     let outer = '';
 
-    // If we're dealing with keyframes just flatten them
-    if (/^@[k|f]/.test(wrapper)) {
-        // Return the wrapper, which should be the @keyframes selector
-        // and stringify the obj which should be just flatten
-        return (
-            wrapper +
-            JSON.stringify(obj)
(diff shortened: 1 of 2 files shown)


src/core/parse.js
```diff
@@ -34,10 +21,22 @@ export const parse = (obj, paren, wrapper) => {
...
+            if (key[0] == '@') {
+                next = paren;
+                // If this is the case for `@font-face`
+                if (key[1] == 'f') {
+                    next = key;
+                }
+            }
+
+            // If this is the `@keyframes`
+            if (/@k/.test(key)) {
```
cristianbote: Just seeing this, make me wonder if I can inline them even more 🤔 cause that would make the `parse` call for `@` rules really specific.

```js
if (key[0] == '@') {
    if (key[1] == 'f') {
    }
    else if (key[1] == 'k') {
  // and so on
    } else {
       blocks += parse(val, next, key);
    }
} else {
    blocks += parse(val, next, wrapper || '');
}
```

Issue #80, opened 2020-01-03
TypeError when using with TypeScript


alexieyizhe:
Hi there! Amazing work with this library.

I'm using it with Preact's TypeScript template, and it's throwing an error since the type expects children to be a property on the goober `StyledVNode`:

![image](https://user-images.githubusercontent.com/17508679/71754262-6d5a0200-2e53-11ea-88be-65a9ad1fe644.png)

When I change the definition to:
```js
export const Text = styled<{ children: any }>("span")`
  color: lightblue;
`;
```
it works fine.

The examples don't seem to have to specify this, so is this something with my code or the library?

Here are my relevant dependencies:
```
    "goober": "^1.6.1",
    "preact": "^10.0.0",
    "typescript": "^3.7.4"
```

Much appreciated!

alexieyizhe:
@cristianbote I've opened a PR that should fix this issue along with one other.

cristianbote:
Fixed in #85

PR #85, opened 2020-01-13
Fix typings for nested styled components and empty props

alexieyizhe:
This PR fixes two issues:

Currently, TypeScript will incorrectly report `styled(InnerGooberComponent)` as invalid since it expects the type of `tag` to be a string. This will error when nesting components together. We add a Union type to allow nesting components.

As well, related to #80,  TypeScript will complain that the type of the VDom element returned by `styled` is incompatible since it does not have a children prop. From https://github.com/DefinitelyTyped/DefinitelyTyped/pull/32843 it seems like the best way to fix this is to add a children prop to the type.

### DIFF
--- a/goober.d.ts
+++ b/goober.d.ts
@@ -5,7 +5,7 @@ export = goober;
 export as namespace goober;
 
 declare namespace goober {
-    function styled<T = { [prop: string]: any }>(tag: string | Tagged<any>): Tagged<T>;
+    function styled<T = {}>(tag: string | StyledVNode<any>): Tagged<T & { children: any }>;
     function setPragma<T>(val: T): void;
     function extractCss(): string;
     function glob(tag: CSSAttribute | TemplateStringsArray | string): void;
--- a/goober.d.ts
+++ b/goober.d.ts
@@ -5,7 +5,7 @@ export = goober;
 export as namespace goober;
 
 declare namespace goober {
-    function styled<T = {}>(tag: string): Tagged<T>;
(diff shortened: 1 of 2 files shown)


goober.d.ts
```diff
@@ -5,7 +5,7 @@ export = goober;
 export as namespace goober;
 
 declare namespace goober {
-    function styled<T = {}>(tag: string): Tagged<T>;
+    function styled<T = {}>(tag: string | StyledVNode<any>): Tagged<T & { children: any }>;
```
alexieyizhe: From https://github.com/DefinitelyTyped/DefinitelyTyped/pull/32843, they detect if the styled component is a React element. Since this library references React and Preact, should we explicitly be checking if the element is a React element?

ts-tests/test-api.tsx
```diff
@@ -26,6 +26,14 @@ const testStyledCss = () => {
         background: black;
     `;
 
+    const EmptyPropsText = styled('p')`
```
alexieyizhe: These components would have previously failed the `test-ts` script.

goober.d.ts
cristianbote: Thank you! 🙏 And I think it's a good call to not tie `goober` to either of them.

ts-tests/test-api.tsx
cristianbote: Really loving this!

cristianbote:
Awesome job @alexieyizhe! Will push a new patch release asap.

Issue #69, opened 2019-12-10
(feat) Add/create a babel plugin for `styled.<tagName>`


cristianbote:
As @jviide built this https://github.com/jviide/babel-plugin-transform-styledodo we could leverage it and create a babel plugin, to help people used with the `styled.div` pattern.

JoviDeCroock:
I don't mind looking into this in the near future, the goal of this would be to achieve syntactical parity to `styled-components or @emotion/styled`?

cristianbote:
Thank you @JoviDeCroock! :) So happy to have you here!

Do you know if there _are_ differences between them?

jviide:
Had a chat with @JoviDeCroock, who encouraged me to finally collect the stuff from https://github.com/jviide/babel-plugin-transform-styledodo and create PR #86.

cristianbote:
Merged in #86

PR #86, opened 2020-01-15
Add a Babel plugin for styled.<tagName> syntax (resembling styled-components)

jviide:
This pull request is based on issue #69 and chat discussions with @cristianbote and @JoviDeCroock. It adds a Babel plugin called babel-plugin-transform-goober under `/packages/babel-plugin-transform-goober`. The plugin is for rewriting [styled-components](https://styled-components.com/)-like `styled.<tagName>` syntax to `styled("<tagName>")` calls.

The plugin rewrites all expressions in the form `styled.*` to `styled("*")`. Therefore things like calls and tagged expressions also work. For example the following file:

```js
style.div;
style.div`...`;
style.div("...");
```

gets rewritten to:

```js
style("div";
style("div")`...`;
style("div")("...");
```

The plugin is enabled by adding it to the Babel configuration:
```json
{
  "presets": [...],
  "plugins": ["babel-plugin-transform-goober]
}
```

By default the plugin assumes that the root identifier is `styled`. If there is a need to use some other name then the plugin can be configured by setting the `"name"` option. For example the configuration:

```json
{
  "presets": [...],
  "plugins": [["babel-plugin-transform-goober", { "name" : "goober" }]]
}
```

allows you to write code like this:

```jsx
import React from 'react';
import { styled as goober } from 'goober';

const Button = goober.button`...`;
```

Related tests are included.

### DIFF
--- a/packages/babel-plugin-transform-goober/README.md
+++ b/packages/babel-plugin-transform-goober/README.md
@@ -1,0 +1,53 @@
+# babel-plugin-transform-goober
+
+A [Babel](https://babeljs.io/) plugin for
+[🥜goober](https://github.com/cristianbote/goober), rewriting `styled.div` syntax to `styled('div')` calls.
+
+## Install
+
+`npm install --save babel-plugin-transform-goober`
+
+## How to use
+
+Edit `.babelrc.json`
+
+```json
+{
+  "presets": [...],
+  "plugins": ["babel-plugin-transform-goober]
+}
+```
+
+And now you can create your components using `styled.*` syntax:
+
+```jsx
+import React from 'react';
+import { styled } from 'goober';
+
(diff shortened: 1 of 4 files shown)


jviide:
Added commit a37efad that modifies the logic a bit. Originally the plugin required the tag name to be an identifier, so `styled.div` did get transformed but `styled["div"]` did not. However, in styled-components `styled["d" + "i" + "v"]` should technically work (?), so a37efad relaxes requirements a bit, and things like `styled["d" + "i" + "v"]` get transformed to `styled("d" + "i" + "v")`.

If this is deemed a bad idea then the commit can just be dropped 🙂

cristianbote:
travis fails because the master was broken by me

PR #94, opened 2020-01-22
goober babel-macro

hadeeb:
A babel-macro to transform `styled.div` to `styled("div")`.

This is my first time doing anything with babel.

I got it working here
[Edit jolly-lehmann-euby4](https://codesandbox.io/s/jolly-lehmann-euby4?fontsize=14&hidenavigation=1&theme=dark)


cristianbote:
Hey @hadeeb 👋 Thanks for opening this PR! Really appreciate it!

To be honest, I haven't used babel-macro yet 😃 is it a choice that users take on here? Is it easier to include a macro rather than a plugin?

hadeeb:
> is it a choice that users take on here?

Yes. Using a plugin is implicit.
Macros are explicit. Every import will have to be changed.

> Is it easier to include a macro rather than a plugin?

Users will have to setup `babel-plugin-macros`. But it's a one time process.
Build tools like create-react-app & Gatsby comes preloaded with it.
macros are the only way to add custom babel behaviours in create-react-app.

The goal is something like babel macros of [styled-components](https://styled-components.com/docs/tooling#babel-macro) and [emotion](https://emotion.sh/docs/babel-macros).

cristianbote:
@hadeeb thank you for your follow-up and explanation! Indeed that'll be awesome to have it.

Do you think we could rename it maybe? `babel-macro-goober`? This goes for folder and package name as well.

cristianbote:
This is great! Thank you for taking the time to do it! Small question: is there a way to have some tests? Something that will make sure this will work fine going forward.

cristianbote:
Published the macro!

hadeeb:
I did have some tests working locally. I couldn't get them working here. Probably some configuration issue. I'll open a PR once I figure it out.

It's available [here](https://github.com/hadeeb/goober.macro/commit/59cdd57b954c0d6df1dd08e9020d49807508d54a).

Issue #89, opened 2020-01-15
css(taggedTemplate) styntax is returning a string instead of a function


raulgg:
I was following the documentation for the css syntax and trying to test it on my project I found that css is returning a string with the classname hash instead of returning a function as showed in the docs.

https://github.com/cristianbote/goober/blob/master/README.md#csstaggedtemplate

I took a look into the source code and it seems that returning the hash instead of the function seems intentional. So I'm guessing that this is matter of updating the docs? Or am I missing something?

cristianbote:
Hey @raulgg! Thanks for opening this issue.

Yeah it's a documentation error 😅 would you like to open a PR to fix it? I would really appreciate it.

raulgg:
Sure, I'll take some time to update the docs over the next two days.

cristianbote:
Closed via @raulgg PR. Thanks a bunch!

PR #95, opened 2020-01-23
Update TypeScript definitions and css API docs

raulgg:
On this change I updated the docs to clarify how the css(taggedTemplate) examples (#89).

I also updated the TypeScript definitions to fix some of the issues we were facing.


cristianbote:
Published in `goober@1.6.5`

PR #115, opened 2020-02-19
🏌️‍♂️ Golf/put put

cristianbote:


src/core/parse.js
```diff
@@ -44,7 +42,7 @@ export const parse = (obj, paren, wrapper) => {
             // Push the line for this property
             else {
                 current +=
-                    (prefixer && prefixer(key.replace(/[A-Z]/g, '-$&').toLowerCase(), val)) ||
+                    (parse.p && parse.p(key.replace(/[A-Z]/g, '-$&').toLowerCase(), val)) ||
```
cristianbote: @JoviDeCroock what do you think? This would shave around `5B` I think

src/styled.js
```diff
@@ -14,15 +18,23 @@ const setup = (pragma, prefix, forward, theme) => {
...
+
+            // Set a flag if the current components had a previous className
+            // similar to goober. This is the append/prepend flag
             _ctx.o = /\s*go[0-9]+/g.test(_ctx.p.className);
+
+            // Define the new className
             return h(
                 tag,
-                Object.assign({}, props, {
+                Object.assign(_ctx.p, {
```
JoviDeCroock: Ctx.p potentially has a theme in it so we would switch the props reference. This doesn’t seem ideal/deterministic

cristianbote: That's a good catch. Thank you! Size is the same 😃

cristianbote:
Huh, this one doesn't really shaves bytes... 😢 gonna look into it

Issue #97, opened 2020-01-24
Planning to support CSSStyleSheet API / insertRule?


manuschillerdev:
Are you planning to support CSSStyleSheet API / insertRule in the future?

If so, i would gladly contribute the feature :) 
I created a minimalistic POC here: https://codesandbox.io/s/rough-platform-vs3t3

Browser Support:
https://caniuse.com/#search=CssStyleSheet

pros:
- runtime performance(!!!)
- maybe we will need some less bytes

cons: 
- styles are not editable via DevTools

emotion uses constructable StyleSheets in prod environments and falls back to regular `<style>` elements in dev environments for a better dev experience (devtools). Maybe that would be an approach for us as well?

cristianbote:
Hey @manuschillerdev Thanks for opening this issue! 😄 Yes! That's on my todo list for while now and didn't had enough time to get to it.

That's a nice demo you got in there! :) seems like it could trim down the sizes as well.

Feel free to open a PR if you want this is awesome!

cristianbote:
I did played around with this one and I have not seen any perf gains from it. 😞  I'll try to give it a go once more but if the gains are not there I think I will drop it for now.

PR #122, opened 2020-03-14
(feat) Change the parser output from concatenated strings to arrays and adopt insertRule

cristianbote:
This is gonna introduce most likely breaking changes, in preparing for v2. 

Todos:
- [ ] Adapt tests
- [ ] Document breaking changes
- [ ] Add a `/global` entry point (since the glob export is gonna be removed)
- [ ] Golf some more

Related: #97

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -125,17 +125,77 @@
       }
     },
     "@babel/helper-create-class-features-plugin": {
-      "version": "7.8.3",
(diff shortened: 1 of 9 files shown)


cristianbote:
I'm gonna park this for now. `insertRule` seems to add a lot of memory overhead, and we would need to keep track ourselves of the rules inserted so we don't duplicate them. That seems to be struggle.

PR #131, opened 2020-04-16
Improve typescript typings

lucat1:
As discussed over slack, I applied a bunch of changes to the `styled` function to improve its type-safety in typescript projects:

- Handle various call types of the function in separate types (as [emotion](https://github.com/emotion-js/emotion/blob/master/packages/styled-base/types/index.d.ts#L138) does)

- Strictly type `styled('<html element>')` calls. Calling `styled('button')` for example returns a typed JSX element with the allowed props and their type value

- Moved nonstandard assumptions of vdom libraries internals(such as `{ children }` to be contained inside the props) to the typescript standard `JSX` namespace, which is implemented from all major libraries such as `preact` and `react`. You can read more about it [here](https://www.typescriptlang.org/docs/handbook/jsx.html)

- Changed types to allow extending existing custom `component`s, either class-based or functional. This implementation also infers types from the parent element, as expected.

- Extended typescript tests and checked that autocompletion works in all components as expected.

- Added `@types/react` as a provider for the JSX namespace, but also other libraries like `preact` would have worked just fine as they implement it too. ([proof](https://github.com/preactjs/preact/blob/master/src/jsx.d.ts))


lucat1:
I would also like to implement types for calling `styled.div` instead of `styled('div')` for people who use both typescript and the babel plugin(like me). Let me know what you think, it would be pretty quick and similar to [emotion's implementation](https://github.com/emotion-js/emotion/blob/master/packages/styled/types/index.d.ts#L24-L172)

lucat1:
Screenshot 2020-04-22 at 17 04 29

That use case is indeed covered as I haven't changed much of the typings for the `styled` function itself. As you can see in the screenshot it also properly picks up the props without any declaration. I've now added yet another typing test so that this use case doesn't ever fall trough.

lucat1:
The only other that could be implemented would be the `styled.div` syntax, which I haven't yet done yet as I'm awaiting your feedback on whether you want this feature to be included.

cristianbote:
Thank you @lucat1! Much appreciated! 🎉

PR #135, opened 2020-04-24
(chore) Fix the perf bottleneck and prepare for v2

cristianbote:


src/core/compile.js
```diff
@@ -8,7 +8,7 @@ export const compile = (str, defs, data) => {
         let tail = defs[i];
 
         // If this is a function we need to:
-        try {
+        if (tail && tail.call) {
```
cristianbote, line 11: This was the _real_ perf hit.

PR #140, opened 2020-05-05
(chore) Revamp the fwd ref and add nested test

cristianbote:
Adds the test for #47 and rewires the forward ref.

@lucat1 pointed out that the current forward ref implementation does not properly work if one tries to import a component before calling `setup`. Which makes total sense. So, this PR address it like this:
* ~~We have support for `styled` to be directly passed to `React.forwardRef`~~ We still have that support but `forwardRef` should be passed as second argument to `styled`.
* `setup` no longer needs the forward ref function.

This boils down to this:
```js
const Foo = styled('div', React.forwardRef)`
    background: tomato;
`;
```


lucat1:
I'd suggest adding some rendering tests with react as I've had errors(at least in with the development mode of react) when I tried to implement this in my own fork.

PR #157, opened 2020-07-09
Fix/remove theme prop from attributes

cristianbote:
There was this annoying bug/side-effect that `theme` would appear as attribute on the rendered html element. It shouldn't be there in the first place. This makes sure that the `style` function will have `props.theme` but not the `h` call.


cristianbote:
Gah! Too much size increase 😞

PR #164, opened 2020-07-18
Enhance babel plugin for nicer developer UX

marvinhagemeister:
This PR adds several features to the `babel-plugin-transform-goober` plugin.

## `#__PURE__`-Annotations

Minifiers have a hard time knowing if a `styled()` call has a side-effect or not. To avoid generating invalid code they side with caution and won't remove unused components.

```js
// Final bundle, Foo is still there despite it not being used anywhere
const Foo = styled("div")``;

export function Bar() {
  return <div />
}
```

By adding a `#__PURE__` comment just before the call expression we can signal minifiers that the next statement has no side effects.

```js
// Before minification
const Foo = /*#__PURE__*/ styled("div")``;

export function Bar() {
  return <div />
}
```

After:

```js
// Final bundle, Foo is gone
export function Bar() {
  return <div />
}
```

This is enabled by default as I can hardly come up with a scenario where one wouldn't want this behavior. The user can opt out by setting `pure: false`.

## Richer component names in Preact devtools

Currently the name of a component will be inferred in this order for function and class components:

1. `fnOrClass.displayName`
2. `fnOrClass.name`
3. `Unknown`

The minified function name that's published on npm is called `a` and therefore every goober component would show up as `a` inside the devtools. By supplying a richer `displayName` the component names become less cryptic in the devtools.

Before:

Screenshot from 2020-07-18 18-14-13

After:

Screenshot from 2020-07-18 18-12-56

This can be enabled by passing `displayName: true` to the babel plugin options.

## Richer CSS class names

While it's nice to have richer names in Preact Devtools, I find myself spending more time with the native elements panel when I'm focusing on the design. So it'd be nice to see which node is coming from which component. To do that the plugin will now add additional CSS classes that show where a component is coming from. They don't set any styles or anything. This is best illustrated with a quick example:

```js
// Before
<div className="go12312323" />

// After
<div className="Button__Icon-go-2 go12312323" />
```

The format for the generated className is `[FILENAME]__[COMPONENT]-go-[#]`.

This can be enabled by passing `displayName: true` to the babel plugin options.

### DIFF
--- a/packages/babel-plugin-transform-goober/index.js
+++ b/packages/babel-plugin-transform-goober/index.js
@@ -1,3 +1,5 @@
+const nodePath = require('path');
+
 /**
  * Prepend #__PURE__ comment to help minifiers with
  * dead code elminiation (=DCE)
@@ -10,13 +12,17 @@ function prependPureComment(node) {
     });
 }
 
+function createAssignment(t, left, right) {
+    return t.expressionStatement(t.assignmentExpression('=', left, right));
+}
+
 module.exports = function ({ types: t }, options = {}) {
     const name = options.name || 'styled';
 
     return {
         name: 'transform-goober',
         visitor: {
-            TaggedTemplateExpression(path) {
+            TaggedTemplateExpression(path, state) {
                 if (t.isIdentifier(path.node.tag) && path.node.tag.name === 'css') {
                     if (options.pure) {
                         prependPureComment(path.node);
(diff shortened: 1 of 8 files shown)


packages/babel-plugin-transform-goober/index.js
```diff
@@ -1,9 +1,86 @@
...
+                        const variable = path.findParent((path) => path.isVariableDeclaration());
+                        if (variable && variable.node.declarations.length === 1) {
+                            const decl = variable.node.declarations[0];
+
+                            if (t.isIdentifier(decl.id)) {
+                                // Add displayName to goober components for easier debugging
+                                const displayName = variable.insertAfter(
+                                    createAssignment(
+                                        t,
+                                        t.MemberExpression(decl.id, t.identifier('displayName')),
```
cristianbote, line 50: Question, is it possible to change `displayName` into something else? Right now `Styled.displayName` adds like `+7B` which are coming from `display` I think. So my thinking is that could we change this to `name` and still be picked-up by devtools and such?

cristianbote: Actually, nope. It's not the `displayName` :D I'll be back. Chasing the `+7B` addition.

marvinhagemeister: Another key won't work I'm afraid :S Setting the function name is a noop in most browsers and will throw in older ones. That's why we use `displayName` in `preact/compat` and it's the same property the react ecosystem uses for the same reasons.

PR #165, opened 2020-07-19
Add support for parsing multiline CSS comments

marvinhagemeister:
This PR adds support for parsing multiline comments inside the CSS string. Previously they were not stripped and would trip up our parser, leading to invalid CSS.

### DIFF
--- a/src/core/astish.js
+++ b/src/core/astish.js
@@ -1,5 +1,5 @@
 let newRule = /(?:([a-z0-9-%@]+) *:? *([^{;]+?);|([^;}{]*?) *{)|(})/gi;
-let ruleClean = /\/\*.*?\*\/|\s{2,}|\n/gm;
+let ruleClean = /\/\*+(?:.|[\n])*?\*\/|\s{2,}|\n/gm;
 
 /**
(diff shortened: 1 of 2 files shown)


src/core/astish.js
```diff
@@ -1,5 +1,5 @@
 let newRule = /(?:([a-z0-9-%@]+) *:? *([^{;]+?);|([^;}{]*?) *{)|(})/gi;
-let ruleClean = /\/\*.*?\*\/|\s{2,}|\n/gm;
+let ruleClean = /\/\*+(?:.|\n)*?\*\/|\s{2,}|\n/gm;
```
cristianbote, line 2: I was playing around trying to bring the added bytes down and I think by having a range with `\s` and `\S` will capture that nasty newline comment. Cause apparently `.`(dot) does not capture newlines, lol.

```suggestion
let ruleClean = /\/\*[\s\S]*?\*\/|\s{2,}|\n/gm;
```

This one adds `+2B` to master. Does it make sense?

marvinhagemeister: Good catch, this is a lot shorter :+1: :100:

PR #166, opened 2020-07-19
Add support for parsing sibling selectors

marvinhagemeister:
This PR adds support for parsing sibling selectors:

```css
.foo {
  :hover,
  :active,
  :focus {
      color: red;

      span {
          color: blue;
      }
  }
}
```

### DIFF
--- a/src/core/parse.js
+++ b/src/core/parse.js
@@ -35,7 +35,9 @@ export let parse = (obj, paren, wrapper) => {
                 blocks += key + '{' + parse(val, '', '') + '}';
             } else {
                 // Call the parse for this block
-                blocks += parse(val, next, next == paren ? key : wrapper || '');
+                next.split(/,/g).some((sib) => {
+                    blocks += parse(val, sib, sib == paren ? key : wrapper || '');
+                });
             }
         } else {
             if (/^@i/.test(key)) {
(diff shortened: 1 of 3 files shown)


marvinhagemeister:
I guess this slightly passes the 1kb threshold :grimacing:

src/core/parse.js
```diff
@@ -35,7 +35,9 @@ export let parse = (obj, paren, wrapper) => {
                 blocks += key + '{' + parse(val, '', '') + '}';
             } else {
                 // Call the parse for this block
-                blocks += parse(val, next, next == paren ? key : wrapper || '');
+                next.split(/,/g).some((sib) => {
+                    blocks += parse(val, sib, sib == paren ? key : wrapper || '');
+                });
```
cristianbote, line 40: 🎉  awesome! 😃 

Was playing around to bring this down, and locally this version
```suggestion
                blocks += next.replace(/([^,])+/g, (group) =>
                    parse(val, group, group == paren ? key : wrapper || '')
                );
```

Adds like `+12B`. What do you think?

marvinhagemeister: Tried out the suggestion but somehow the test fails. Need to check what's going on 👍

cristianbote:
Already merged in #172

PR #168, opened 2020-07-19
Simplify styled test cases to use less mocking

marvinhagemeister:
This PR rewrites the styled test suite to rely less on heavy mocking of internals and instead they're only verifying the end result. They're basically rewritten to be proper integration tests, which should make it easier for us to experiment with internals against a stable test suite that doesn't break when we rename an internal function.

The changes here are arguably subjective. For me this makes the tests easier to follow along and they're a little shorter to write. Let me know what you think :+1: 

(The inline diff view in GitHub is a little confusing. The differences are easier to see when compared side by side)

### DIFF
--- a/src/__tests__/styled.test.js
+++ b/src/__tests__/styled.test.js
@@ -1,166 +1,139 @@
-import { h, render } from 'preact';
 import { styled, setup } from '../styled';
-import { css } from '../css';
+import { extractCss } from '../core/update';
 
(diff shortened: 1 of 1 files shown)


cristianbote:
@marvinhagemeister should we merge this or wait for the other ones first?

marvinhagemeister:
@cristianbote I think this one is ready to be merged. The others don't touch the same test file 👍

PR #170, opened 2020-07-24
(feat) Drop theme prop

cristianbote:


cristianbote:
Really not happy size-wise but, it's really needed 😞 .

Thanks to @marvinhagemeister for the updated tests for `styled` 😃  it was sooo much easier to work with!

Issue #49, opened 2019-05-03
(bug) Nesting selectors with multiple bodies does not get splitted into two separate rules


cristianbote:
As @uppercod stated on twitter(https://twitter.com/Uppercod/status/1118912434239090688). 
```
&:hover,
&:focus {
  border: 0;
  span {
    index: unset;
  }
}

// Should get transpiled to:
&:hover,
&:focus {
  border: 0;
}

&:hover span,
&:focus span {
  index: unset;
}
```

He kindly already has a fix for this #47! 🎉

cristianbote:
Fixed

PR #172, opened 2020-07-27
(chore) parse improvements

cristianbote:
This actually tackles the nested parse improvements from #166 and other issues. Also this one enables the `@support` rule and the rest of the bunch. Test pending.

This kinda adds a few bytes to the size but it's totally worth it. I'll explore more though.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "goober",
-  "version": "2.0.3",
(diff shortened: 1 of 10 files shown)


cristianbote:
This will fix #49

cristianbote:
@marvinhagemeister how do you feel about this one? Basically it does what yours does but also tries to have the smallest footprint.

PR #177, opened 2020-09-07
add // style comments

russelgal:


cristianbote:
Heyo! Thank you for opening this one!

Would you mind adding a test case as well? Also I'm looking at the added size, +36B any chance we could do it in less?

cristianbote:
Hey @russelgal, still interested in pursuing this one? It needs a test case for this -- just so we'll be able to track it along the way. Thank you so much for your time!

PR #181, opened 2020-09-25
(feat) keyframes

cristianbote:
This one adds support for `keyframes` function, same thing as `glob` but, this one hashes the animation names so one could reuse the same animations if wanted.

Could be used in all sorts of ways, as an example:
```js
import { keyframes, css } from 'goober';

const fade = keyframes`
  to {
    opacity: 1;
  }
`;

const cls = css`
  opacity: 0.5;

  &:hover {
    animation: 1s ${fade} ease-out;
  }
`;
```

Still need to sort out:
- [x] Tests
- [ ] Size shavings
- [x] Update documentation
- [ ] Think about moving `glob` and `keyframes` into separate entry points. So they won't affect the core size but I am not sure about it.

### DIFF
--- a/src/core/hash.js
+++ b/src/core/hash.js
@@ -17,7 +17,7 @@ let stringify = (data) => {
     let out = '';
 
     for (let p in data) {
-        const val = data[p];
+        let val = data[p];
         out += p + (typeof val == 'object' ? stringify(data[p]) : data[p]);
     }
 
--- a/src/core/hash.js
+++ b/src/core/hash.js
@@ -34,21 +34,24 @@ let stringify = (data) => {
  * @returns {String}
  */
 export let hash = (compiled, sheet, global, append, keyframes) => {
-    let ast = compiled[0] ? astish(compiled) : compiled;
-    let stringifiedCompiled = stringify(ast);
+    let stringifiedCompiled = typeof compiled == 'object' ? stringify(compiled) : compiled;
     let className =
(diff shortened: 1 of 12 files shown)


src/core/hash.js
```diff
@@ -30,21 +30,22 @@ let stringify = (data) => {
...
  * @param {Boolean} append Append or not
+ * @param {Boolean} keyframes Keyframes mode. The input is the keyframes body that needs to be wrapped.
  * @returns {String}
  */
-export let hash = (compiled, sheet, global, append) => {
-    // generate hash
-    let stringifiedCompiled = compiled.toLowerCase ? compiled : stringify(compiled);
+export let hash = (compiled, sheet, global, append, keyframes) => {
+    let ast = compiled[0] ? astish(compiled) : compiled;
+    let stringifiedCompiled = stringify(ast);
```
cristianbote, line 38: I notice a pref deopt and I think this is the reason. `stringify`-ing all the time. Will investigate.

cristianbote:
Really not happy size-wise but the more I had to work at this the more I fell in love with it! 😄 I can actually see me using it quite often. Thanks @dvoytenko for opening up the issue about it. Much appreciate it!

dvoytenko:
Thanks a lot, @cristianbote! I'll try it out very soon as well. The binary-size: I'd assume it's eliminable?

dvoytenko:
Just tried it out and it works very well! Thanks again, @cristianbote!

PR #189, opened 2020-10-22
docs: extend setup api with theme example

devrnt:
Add an example in the `README` to showcase the `setup` with a theme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,6 +203,22 @@ import { setup } from 'goober';
 setup(React.createElement);
 ```
 
+With theme:
+```js
+import React from 'react';
+import { setup, styled } from 'goober';
+
+const theme = { primary: 'blue' };
+const ThemeContext = createContext(theme);
+const useTheme = () => useContext(ThemeContext);
+
+setup(React.createElement, undefined, useTheme);
+
+const ContainerWithTheme = styled('div')`
+  color: ${(props) => props.theme.primary};
+`;
+```
+
 ### `css(taggedTemplate)`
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -203,6 +203,22 @@ import { setup } from 'goober';
...
+const ThemeContext = createContext(theme);
+const useTheme = () => useContext(ThemeContext);
+
+setup(React.createElement, undefined, useTheme);
+
+const ContainerWithTheme = styled('div')`
+  color: ${(props) => props.theme.primary};
+`;
+```
+
```
cristianbote, line 221: This is really awesome @devrnt! Thank you for opening this one and fix it! 🎉 

Was about to suggest do add the `With theme` as a nested title so it shows in the TOC at the top, but then notice that the `With prefixer` section is missing as well. Do you wanna tackle the with prefixer one as well? This is totally approved by the way! 👍

cristianbote (approved):

README.md
devrnt: I'll open up a new PR with the nested `With theme` title. Can you point to the `With prefixer` section, I can't seem to find what you're talking about 🤔

cristianbote: Awesome!

Yeah, sorry if the above is confusing. I've opened #191 for it as well. So your entry for the _theme_ param is really awesome, but that made me realise that another for _prefixer_ does not exist and it should. Inside here https://github.com/cristianbote/goober#setuppragma-function-prefixer-function-theme-function.

PR #192, opened 2020-10-22
docs: extend setup api with prefixer example

devrnt:
Add an example in the README to showcase the setup with a prefixer.


cristianbote (approved):

cristianbote:
@devrnt you are awesome 👍 thank you for this!👏

PR #193, opened 2020-10-24
Add support for theme TypeScript definitions

devrnt:
By letting the user extend the `goober` TypeScript definitions it is possible to support theme types. Should be used in combination when using theme in `setup(X, X, theme)`.

It works as follows:
1) The user should create a declarations file, let's call it `goober.d.ts` (file name suggestions?)
```ts
import 'goober';

declare module 'goober' {
    export interface DefaultTheme {
        colors: {
            primary: string;
        };
    }
}
```
2) Use it like you should normally do (but with "autocomplete")
```js
const ThemeContainer = styled('div')`
     background-color: ${(props) => props.theme.colors.primary};
`;
```
`DefaultTheme` is being used as an interface of `props.theme` out of the box.

If this PR gets approved, I think this should be moved to a section in the `README`.

Inspiration: https://styled-components.com/docs/api#create-a-declarations-file

### DIFF
--- a/goober.d.ts
+++ b/goober.d.ts
@@ -5,10 +5,14 @@ export = goober;
 export as namespace goober;
 
 declare namespace goober {
+    interface DefaultTheme {}
+
+    type Theme<T extends object> = keyof T extends never ? T : { theme: T };
+
     interface StyledFunction {
         // used when creating a styled component from a native HTML element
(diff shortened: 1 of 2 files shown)


cristianbote (approved):
Awesome stuff! @devrnt Thank you 🙏

cristianbote:
Published `2.0.11`

PR #141, opened 2020-05-05
Support node 13

nsivertsen:
Hi there! First of all, thanks for this wonderful library. I'm really enjoying using it.

I've been trying to use it with Node 13's new ESM support, and saw that that was still missing, hence this PR. I've basically copied exactly what Preact is doing: https://github.com/preactjs/preact/pull/2451.

Of the sub-packages, only `goober-autoprefixer` is currently being built, as it appears the other ones don't need to be built at all, as far as I can see?

I've added a basic test, taken directly from HTM: https://github.com/developit/htm/pull/164

The filesize test is currently failing, which is why this might fail in CI.

PR #195, opened 2020-10-27
fix package entry points

osdevisnot:
I was trying to use `goober` with `snowpack` when I started noticing [this error](https://github.com/snowpackjs/snowpack/discussions/1421).

As mentioned in [specs](https://nodejs.org/api/packages.html#packages_package_entry_points), It seems, when `exports` is present, we can not use `main` as a fallback.

This PR should fix the usage for snowpack and modern node versions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,10 @@
   ],
   "exports": {
     ".": {
-      "import": "./dist/goober.module.js"
+      "browser": "./dist/goober.module.js",
+      "import": "./dist/goober.module.js",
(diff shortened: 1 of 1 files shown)


osdevisnot:
Please note that the auto prefixer will be fixed in #141

cristianbote:
Oh that's a great catch @osdevisnot! Thank you so much for opening this up. So many specs to be complaint with 😄 but we'll get there. Will merge and publish a new patch. Will comment here when it's done.

PR #198, opened 2020-11-09
(feat) Conditional styles with array as argument

cristianbote:
This is a new way, for goober, to define your conditional styling based on an array of objects or array of functions.

cc @ben-rogerson

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,6 +194,20 @@ const Btn = styled('button')((props) => ({
 <Btn size={20} />;
 ```
 
+##### Arrays
+
+```js
+import { styled } from 'goober';
+
+const Btn = styled('button')([
+    { color: 'tomato' },
+    ({ isPrimary }) => ({ background: isPrimary ? 'cyan' : 'gray' })
+]);
+
+<Btn />; // This will render the `Button` with `background: gray;`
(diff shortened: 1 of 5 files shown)


ben-rogerson:
Thanks so much for adding this feature. I've tested it's all working great!

I also tested with `babel-plugin-transform-goober` but to add [twin](https://github.com/ben-rogerson/twin.macro) support the plugin must allow working with a renamed import from styled (eg: `import { styled as renamedImport } from 'goober'`). 
I have that feature working locally and I'll submit a pr shortly.

ben-rogerson:
I found a small issue:

image

The error shows when the array item equals `undefined`, like this:

```js
const Button = styled('button')(({ isPrimary }) => [
  isPrimary && {
    backgroundColor: 'green'
  }
])

const App = () => (
    <Button>Send</Button>
)
```

Perhaps you could check if `i` is defined before the merge here?

```js
_val.reduce((o, i) => Object.assign(o, i.call ? i(ctx.p) : i))
```

cristianbote:
@ben-rogerson great catch! Let me add that suggestion.

cristianbote:
Published `2.0.14` with this

ben-rogerson:
Found an issue, possibly related #199

PR #200, opened 2020-11-12
(fix) Make sure reduce works on a new object

### LINKED ISSUE #199: Last class reused because of common sub styles
While testing the recent changes to Goober I came across a strange bug.

The class of an element gets reused for another because they share a group of styles.
The common styles are shown here, named `stylesBase`:

```js
const stylesBase = { padding: '20px', color: 'white' }

// This is green, looks like the class is being reused
// for both the 'Button' and the 'div' components
const Button = styled('button')(stylesBase)

const App = ({ isPrimary = true }) => (
  <div>
    <Button>This shouldn't be green</Button>

    <div
      className={css([
        stylesBase,
        isPrimary && { 'background-color': 'green' }
      ])}
    >
      a div
    </div>
  </div>
)
```

Take a look at [the demo on codesandbox](https://codesandbox.io/s/goober-issue-example-jyed1?file=/src/index.js)

cristianbote:
Fixes #199

### DIFF
--- a/src/__tests__/integrations.test.js
+++ b/src/__tests__/integrations.test.js
@@ -1,6 +1,6 @@
 import { h, createContext, render } from 'preact';
 import { useContext, forwardRef } from 'preact/compat';
-import { setup, styled, keyframes } from '../index';
+import { setup, css, styled, keyframes } from '../index';
 import { extractCss } from '../core/update';
 
 describe('integrations', () => {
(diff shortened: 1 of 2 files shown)


ben-rogerson:
Nice one, you're on 🔥  with these fixes.

I'm surprised with how helpful the codesandbox integration is here too.

cristianbote:
Oh thank you! 😄  I wouldn't know about it if you wouldn't open the issue, so, a big thank you!

About CSB CI, right?! 😄  I feel more at ease that I can test out the examples before committing to anything.

PR #202, opened 2020-11-23
chore: add typescript theme for styled components

devrnt:
Addition to #193 

Add TypeScript theme support for styled components, atm there is only support for native html elements


cristianbote (approved):
Nicely done!

Issue #203, opened 2020-11-24
React does not recognize the XX prop on a DOM element


devrnt:
**Description**
React console error whenever passing props to a goober styled component:
`
React does not recognize the XX prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase XX instead. If you accidentally passed it from a parent component, remove it from the DOM element.
`

**When happening**
Passing through some props to a goober styled component (see codesandbox)

**Possible solution?**
See `transient props` from `styled-components`

**Codesandbox**
See the console logs
https://codesandbox.io/s/vigorous-gates-0o2ge?file=/src/App.js

cristianbote:
Gosh, I really do not want to have to use `$<name>` props naming 🤕  looking into it

devrnt:
Hmm, I currently see no other option than keeping a list of "default" props (don't know if this is desirable).

I made a fork to look at a solution with a `$` prefix and it seems to work. I'll keep this PR on idle until this ticket gets any update

jayalfredprufrock:
Landed here after using the cxs project which had similar goals but is no longer actively maintained. That library solved the problem by automatically stripping properties defined on propTypes (https://github.com/cxs-css/cxs/blob/eb934203a17e07aec63e971c3b87068e911b5b6e/src/component.js#L11) In combination with a babel transform that generates the propTypes from typescript typings, the developer experience was pretty darn good, with the obvious downside of having to ship propTypes in production.

Some might find the small increase in bundle size worth it to not have to prefix every style prop, especially when integrating with something like styled-system. I would also welcome the ability to pass off the burden to the developer via a shouldForwardProp function like what emotion exposes. Or perhaps a babel plugin could be used to generate an optimal black/white list of props so the library wouldn't have to ship with an exhaustive list...idk just thinking out loud. I just know that I've seen the magical prefix approach pitched in numerous style libraries and it always seemed like more people were against it than for it...then again I could be biased on that front :) Either way, this library is looking like an attractive alternative to cxs and I plan on giving it a spin soon!

cristianbote:
Hey @jayalfredprufrock thank you so much for weighing in! Much appreciate it!

After reading your argument and looking into emotion's `shouldForwardProp` functionality and I am really digging it! `goober` currently supports `forwardRef` via the second argument, so the second argument is taken. But, what if you could pass a prop filter to `setup` instead? 🤔 

```js
setup(h, prefix, theme, shouldForwardRef);

[...]

PR #210, opened 2020-12-08
feat: add transient props

devrnt:
Proposal for #203 

To be clear this `$` prefix could be anything even the 🥜 emoji

If this proposal would be approved there are still some things to be done
* [ ] Add documentation
* [ ] Add tests
* [ ] Add example in `ts-tests/test-api.tsx`


cristianbote:
@devrnt thank you for opening this one! Much appreciate it! Size-wise it adds `+30B` which it's a bit much for goober's constraints, but was thinking if we could merge the `Object.assign` call and this one in a way that would have minimal impact?

devrnt:
@cristianbote I don't know how we could move this in the `Object.assign` call.

Funny to see that only 30 bytes are resulting in a +2,65% size change. This small size of goober is just awesome, congrats!

cristianbote:
Yeah it's a blessing but also a curse 😄  lol.

Regarding the transient props. Do you think we could have a test that would check for their integration? Meaning that `React` will complain or something like that?

devrnt:
I don't know how you would intercept React error logs in a test environment.

cristianbote:
@devrnt #212 will shave some `-12B` from the bundle. 😄  do you have ideas if we could achieve it that? 🤔

devrnt:
@cristianbote I don't understand your sentence? You mean like shaving a few more bytes in this PR?

dqbd:
There might be some spare change in terms of file size optimalization (replacing const with let, defining object right at the top), but I suppose every byte counts? 😄 :

```
let _props = Object.assign(props), _previousProps = {};

...

for (let key in _props) {
    if (key[0] !== "$") {
        _previousProps[key] = _props[key]
    } 
}
```

PR #216, opened 2020-12-29
style: use extracted variable

bibidu:


src/core/hash.js
```diff
@@ -18,7 +18,7 @@ let stringify = (data) => {
 
     for (let p in data) {
         let val = data[p];
-        out += p + (typeof val == 'object' ? stringify(data[p]) : data[p]);
+        out += p + (typeof val == 'object' ? stringify(val) : val);
```
cristianbote, line 21: Hey @bibidu thank you for catching this! Unfortunately this ads size to the bundles. Check this output https://github.com/cristianbote/goober/runs/1621077194?check_suite_focus=true#step:3:116

PR #217, opened 2020-12-30
misc: replace Object.assign with spread

dqbd:
We could shrink the library file size by utilizing object spread instead of Object.assign, as I was wondering whether we could fit #210 within the file size constraints.

This does come with a downside of [limited browser support](https://caniuse.com/mdn-javascript_operators_object_initializer_spread_properties), but I suppose that shouldn't be an issue ([Object.assign](https://caniuse.com/mdn-javascript_builtins_object_assign) has marginally better support?), as README.md states that the library works with IE9? Correct me if I'm wrong 😄 

TODO: Performance impact shall be measured, especially around `css.js`, where we are creating a new object in this PR instead of reusing the accumulator.


cristianbote:
Hey @dqbd thank you for opening this! Amazing byte shavings! 🎉  😄 

Related to browser support, I am inclined to move the versions up a bit and if that means smaller footprint by all means, we should totally do it! 👍 

Let me test the perf in a bit and get back.

cristianbote:
Unfortunately this hurts the perf 😞  on node v12.19.1 I get `-3-4k` iterations and on the client there's a `.8ms` degradation. At this point I think we could put a pin on this idea, and revisit later. At least we know that we can shave some more bytes with this.

src/css.js
```diff
@@ -16,7 +16,7 @@ function css(val) {
                 ? // Tagged templates
                   compile(_val, [].slice.call(arguments, 1), ctx.p)
                 : // Regular arrays
-                  _val.reduce((o, i) => (i ? Object.assign(o, i.call ? i(ctx.p) : i) : o), {})
+                  _val.reduce((o, i) => (i ? { ...o, ...(i.call ? i(ctx.p) : i) } : o), {})
```
dqbd, line 19: It seems like as a compromise I would revert this line back, as this would most likely cause a perf drop, what do you think @chrstntdd ?

cristianbote:
Closing due to perf loss. If you think this is doable without perf loss please reopen this. Thank you!

PR #220, opened 2021-01-02
(fix) Bug when extending styles (#218)

### LINKED ISSUE #218: Bug when extending styles
First of all, thanks for this lib, I'm amazed by how much it can do while being less than 1kb.
I have reduced my bundle size by a lot using this with preact.

Unfortunately I have run into a bug while extending styles.
Here is a minimal code sample to reproduce the bug:

```javascript
import React from 'react'
import { render } from 'react-dom'
import { styled, setup } from "goober"

setup(React.createElement)

const RedBase = styled('div')`color: red;`
const Orange = styled(RedBase)`color: orange;`
const Blue = styled(RedBase)`color: blue;`
const UnderlinedBlue = styled(Blue)`text-decoration: underline;`

const App = () => {
	return <div>
		<Orange>Orange</Orange>
		<UnderlinedBlue>UnderlinedBlue</UnderlinedBlue>
	</div>
};

render(<App />, document.getElementById('app'))
```

In this situation the text in UnderlinedBlue should be blue, but it's red because of a wrong order of the class names in the class attribute. In that case, the classname of `RedBase` is after the classname of `Blue` and overwrite it.

However if I remove `<Orange>Orange</Orange>`, it is blue as expected.

I am using goober 2.0.19-next.1
I will try to find some time to fix it.

happy new year !

atersolis:
A proposal to fix #218
I couldn't find anything better
It adds a comment like /*go4214727443*/ before every style in the stylesheet so that you can easily insert other styles right before it.
This way the styles are declared in the right order.
Unfortunately it adds around 30 bytes. 1kb sure is small... It's hard to make some changes without significantly increasing the bundle size.
I hope you can find a better fix.

### DIFF
--- a/src/core/update.js
+++ b/src/core/update.js
@@ -16,6 +16,9 @@ export let extractCss = (target) => {
  * @param {Object} sheet
  * @param {Boolean} append
  */
-export let update = (css, sheet, append) => {
-    sheet.data.indexOf(css) == -1 && (sheet.data = append ? css + sheet.data : sheet.data + css);
+export let update = (css, sheet, before) => {
+    sheet.data.indexOf(css) == -1 &&
+        (sheet.data = before
+            ? sheet.data.replace('/*' + before[0] + '*/', css + '/*' + before[0] + '*/')
+            : sheet.data + css);
 };
--- a/src/styled.js
+++ b/src/styled.js
@@ -35,7 +35,7 @@ function styled(tag, forwardRef) {
             // Set a flag if the current components had a previous className
             // similar to goober. This is the append/prepend flag
             // The _empty_ space compresses better than `\s`
-            _ctx.o = / *go\d+/g.test(_previousClassName);
+            _ctx.o = (_previousClassName || '').match(/ *go\d+/g);
 
             _props.className =
                 // Define the new className
--- a/src/core/hash.js
+++ b/src/core/hash.js
@@ -29,11 +29,11 @@ let stringify = (data) => {
  * @param {String|Object} compiled
  * @param {Object} sheet StyleSheet target
  * @param {Object} global Global flag
- * @param {Boolean} append Append or not
+ * @param {Boolean} before To insert style before an other style
  * @param {Boolean} keyframes Keyframes mode. The input is the keyframes body that needs to be wrapped.
  * @returns {String}
  */
-export let hash = (compiled, sheet, global, append, keyframes) => {
+export let hash = (compiled, sheet, global, before, keyframes) => {
     // Get a string representation of the object or the value that is called 'compiled'
     let stringifiedCompiled = typeof compiled == 'object' ? stringify(compiled) : compiled;
 
@@ -55,7 +55,8 @@ export let hash = (compiled, sheet, global, append, keyframes) => {
     }
 
     // add or update
-    update(cache[className], sheet, append);
+    //"/*" + className + "*/" is a marker to insert styles before .className
+    update('/*' + className + '*/' + cache[className], sheet, before);
 
     // return hash
     return className;


src/core/update.js
```diff
@@ -16,6 +16,9 @@ export let extractCss = (target) => {
  * @param {Object} sheet
  * @param {Boolean} append
  */
-export let update = (css, sheet, append) => {
-    sheet.data.indexOf(css) == -1 && (sheet.data = append ? css + sheet.data : sheet.data + css);
+export let update = (css, sheet, before) => {
+    sheet.data.indexOf(css) == -1 &&
+        (sheet.data = before
+            ? sheet.data.replace('/*' + before[0] + '*/', css + '/*' + before[0] + '*/')
```
cristianbote, line 22: Alrighty, couldn't find time to look into it properly, but now I did. You did an excellent job in here! 😄  really love it!

Do you think we could actually use the bellow instead of adding a comment? 🤔 
```suggestion
            ? sheet.data.replace('.' + before[0], css + '.' + before[0])
```

If we do the above, the changes to `hash.js` are not needed, and that makes the size hit around `+10B` which is really awesome to fix this bug! 🎉

src/styled.js
```diff
@@ -35,7 +35,7 @@ function styled(tag, forwardRef) {
             // Set a flag if the current components had a previous className
             // similar to goober. This is the append/prepend flag
             // The _empty_ space compresses better than `\s`
-            _ctx.o = / *go\d+/g.test(_previousClassName);
+            _ctx.o = (_previousClassName || '').match(/ *go\d+/g);
```
cristianbote, line 38: This one adds `+0B` 😄

src/core/update.js
atersolis: Thanks for taking a look at this.
Well in some cases (like media queries) I think there might be some issues with your suggestion.
Like in this case 

```javascript
const RedBase = styled('div')`
  @media (max-width: 1245px) {
    color: #FF2222;
  }
  @media (max-width: 800px) {
    color: #FF4444;
  }
`;
```

In that case if you try to use `sheet.data.replace('.' + before[0], css + '.' + before[0])` it would be a problem:
```css
@media (max-width: 12450px) {
    /* THE NEW STYLE WOULD BE INSERTED HERE CAUSING A BUG */
    .go2087537148{color:#FF2222;}
}
@media (max-width: 800px) {
    .go2087537148{color:#FF2222;}
}
```

But adding comments before every style definitely feels hacky...
There's probably a better solution, it might be interesting to look at how styled-components or emotion handle this problem.

cristianbote: Oh did not thought about media queries. Can you make me understand why that would bug out? 

Tried it with the changes in this codesandbox https://codesandbox.io/s/vanilla-forked-wprg1?file=/src/index.js:247-390 and to my untrained eye, looks fine? 🤔

atersolis: oh indeed my example does not work

I forked your codesandbox : [https://codesandbox.io/s/vanilla-forked-h06f6?file=/src/index.js](https://codesandbox.io/s/vanilla-forked-h06f6?file=/src/index.js)
You can see here that RedBase is black instead of red, because the class for red is inserted inside the first media-query for the orange style.

cristianbote: Uh oh! Indeed!

So, that's happening because there are no other styles outside the media queries. Do you see this as an established pattern? To have a component solely based on media queries? 🤔

cristianbote:
@atersolis what do you think about this one? I'm really on the fence because of the size increase 😞  do you think having a note about the _known issue_ for the media query above will help? And merge in the replace method?

atersolis:
I don't know if this pattern is common, but it can happen.
Also media queries is only an example, there might be other cases we didn't think about.
I don't think it's worth creating potential bugs just to save a few bytes. It might discourage people from using your lib (it would be a shame). Personally I would prefer a stable 2kb-3kb library over a buggy 1kb one.

riccardogiorato:
We are having this same issue, even if the bundle increase the library won't have bugs like this!

buchansm:
What's the status of this PR? If this bug isn't fixed then `goober` needs to be sunsetted.

cristianbote:
> What's the status of this PR? If this bug isn't fixed then `goober` needs to be sunsetted.

@buchansm what's the status of your financial contribution https://opencollective.com/goober?

Issue #221, opened 2021-01-03
Documentation Site


vandreleal:
Hi @cristianbote,

Congratulations on creating goober! I've created a documentation website and would like to check if you like the idea of having a website for goober. The website was created using [Docusaurus 2](https://v2.docusaurus.io) and is currently hosted at https://gooberjs.github.io. I've also invited you to the organization `gooberjs` that I created for hosting additional resources in the future.

Thanks!

cristianbote:
Wooow!

Thank you @vandreleal! Much appreciate the effort! Amazing! 👍  😃 

Do you think a new organisation for `gooberjs` would be a great fit? I've played around with the idea a while back as well, trying to consolidate the ecosystem, and it felt a bit odd to me.

My reasoning was that for documentation, like you've so amazingly generated the docs 🎉 , a `docs` folder in the main repo would do it. Then for the rest of packages, they could live under the `packages` folder due to tight integration between the main source. Which if they would be to be moved outside the main repo some sort of package alias orchestration should have to happen and I am not that good at it nor I want to spend time on it.

Also this relates to #112.

Would you be open to move the docs and pages into the main repo?

Thank you so much! Waw, I'm blown away by this. Amazing.

vandreleal:
Happy to hear that, @cristianbote! 

Sure thing! I'll fork the repo and play a little with the deployment to see if having the docs as a folder in the main repo would affect the ability to use a GitHub workflow to handle deployments automatically. Get back to you soon!

vandreleal:
Hi @cristianbote, I've just created the PR. The workflow `docs` takes care of deploying the site whenever there's a push to master. After the first successful run, you will need to go to https://github.com/cristianbote/goober/settings and set the GitHub Pages source to `gh-pages`. The website will be available at https://cristianbote.github.io/goober. I'll create a DocSearch key request at https://docsearch.algolia.com/apply after the website is published and update the Algolia section sometime this week.

cristianbote:
Something is up with the way docusaurus has generated the website? 🤔  https://goober.js.org/ looks odd 😄

cristianbote:
Maybe we could buikd it by hand with `wmr`+`goober` based on the root `Readme.md`?

vandreleal:
That's odd., @cristianbote. Have you set the custom domain in the settings page to https://goober.js.org?

vandreleal:
[...]

PR #224, opened 2021-01-04
feat(docs): goober documentation

vandreleal:
https://github.com/cristianbote/goober/issues/221


docs/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2021 gooberjs
```
cristianbote, line 3: Really small nit: I think the project name should be in here without the `js` part. What do you think?

docs/docusaurus.config.js
```diff
@@ -0,0 +1,67 @@
...
+                {
+                    href: 'https://github.com/cristianbote/goober',
+                    label: 'GitHub',
+                    position: 'right'
+                }
+            ]
+        },
+        footer: {
+            style: 'dark',
+            copyright: `Copyright © ${new Date().getFullYear()} goober. Built with Docusaurus.`
```
cristianbote, line 45: 😄

cristianbote:
🙌  tremendous job @vandreleal! This looks awesome!

Do you think that the `README` should point to the docs after this? Feels like the information is doubled? Also I kinda think that having the README available with the docs it's nice.

docs/LICENSE
vandreleal: Good catch! Just removed the `js` part.

vandreleal:
> 🙌 tremendous job @vandreleal! This looks awesome!
> 
> Do you think that the `README` should point to the docs after this? Feels like the information is doubled? Also I kinda think that having the README available with the docs it's nice.

@cristianbote, I think that keeping both of them would be good for those who prefer browsing the `README`, however, there's the small inconvenience of remembering to replicate changes between the `README` and `docs`. The URL to the docs could be pointed using the about repo settings.

PR #227, opened 2021-01-05
feat: add should forward prop

devrnt:
Integration for #203 

Follow up for https://github.com/cristianbote/goober/pull/210 since I accidentally removed my previous repo


src/styled.js
```diff
@@ -46,7 +47,16 @@ function styled(tag, forwardRef) {
...
+            let _propsToPass = forwardProp
+                ? Object.keys(_props)
+                      .filter(forwardProp)
+                      .reduce((acc, curr) => {
+                          acc[curr] = _props[curr];
+                          return acc;
+                      }, {})
+                : _props;
+
+            return h(_props.as || tag, _propsToPass);
```
cristianbote, line 59: Love it! Awesome job @devrnt!

This weighs in at `+38B` 😢 

If we would do something like the bellow, it weighs `+27B` 😞 
```js
            if (fwdProp) {
                for (let p in _props) {
                    if (!fwdProp(p)) {
                        delete _props[p];
                    }
                }
            }

            return h(_props.as || tag, _props);
```

But, what if instead of iterating the props ourselves we let the user do it, adds `+8B` 😀 
```js
            if (fwdProp) fwdProp(_props);

            return h(_props.as || tag, _props);
```
What do you think?

devrnt: Hmm, I see your point to keep this lib as small as possible. But I don't know if this iterating over the props by the user is a good developer experience.  Personally I wouldn't mind, but for new `goober` users it can seem confusing?

We would move  the extra bits to the user's  bundle instead of in this lib, but in a production bundle it'll come to the same thing right? Feels like a zero operation in the end (but I know this lib is expected to be as small as possible).

Either way I would definitely ship your suggestion 🤞

 Thoughts?

cristianbote: Thank you @devrnt. Really great points and you are right. We'll have to see what to shave out from the core instead 😅.

So I guess the one with the for loop then?

devrnt: Sounds perfect to me!

devrnt: Either way I'm thinking if this feature isn't heavily requested we could still introduce the loop over props like you suggested: in https://github.com/cristianbote/goober/issues/203#issuecomment-755203083 instead of the +27B solution.

So the bundle size doesn't get increased that much, it's up to you

cristianbote: Yeah, was initially thinking to suggest adding it as an addon, so would be an opt-in without the size hit 😄 .

```js
import { setup } from 'goober';
import { shouldForwardProp } from 'goober/forward-prop';

setup(pragma, prefix, theme, shouldForwardProp(prop => {
  return prop[0] !== '$';
}));
```

Would it be better? from a DX point? 🤔

cristianbote: Also if we do it as an addon we could also take care of `as` by default.

devrnt: Seems good to me! But you would still have a minor size hit right?

cristianbote: Well `+8B` to core. Then the rest is user land implementation.

cristianbote:
```suggestion
           // Handle the forward props filter if defined
           if (fwdProp) fwdProp(_props);

            return h(_props.as || tag, _props);
```

devrnt: Awesome!

src/__tests__/styled.test.js
```diff
@@ -130,4 +130,16 @@ describe('styled', () => {
         const vnode = Comp({});
         expect(vnode).toMatchVNode('tag', { className: 'go foobar' });
     });
+
+    it('omits css prop with falsy should forward prop function', () => {
+        const shouldForwardProp = (prop) => !prop.includes('$');
```
cristianbote, line 135:
```suggestion
        const shouldForwardProp = (props) => {
            for (let prop in props) {
                if (prop.includes('$')) delete props[prop];
            }
        };
```

cristianbote:
🙏 thank you @devrnt! Solid work in here!

Issue #228, opened 2021-01-06
Preact cli plugin to extract styles


gerhardsletten:
I anyone need to extract styles in a preact cli application I made a plugin which does that.

[gerhardsletten/preact-cli-goober-ssr](https://github.com/gerhardsletten/preact-cli-goober-ssr)

cristianbote:
This is amazing @gerhardsletten! Would you like to open a PR to have it included in the `README`?

gerhardsletten:
@cristianbote Thanks, Goober is also even more amazing, with the joy of writing css-in-js without bundle-bloat ;-)

Maybe it should be added below a `Ecosystem` section?

cristianbote:
@gerhardsletten how do you feel about placing it in here https://github.com/cristianbote/goober#integrations? 🤔

PR #229, opened 2021-01-07
Added info about preact cli plugin

gerhardsletten:
Added PS from #228


cristianbote (approved):
Pure awesomeness! 😄

cristianbote:
OH totally blanked out, but would you mind adding the same inside the docs as well? https://github.com/cristianbote/goober/tree/master/docs/docs/integrations that would mean a lot!

PR #233, opened 2021-01-13
docs: add transient forward prop in readme

devrnt:
Add additional forward prop info in `README`


cristianbote:
🙌 woop woop! This is awesome!

PR #234, opened 2021-01-13
Feat/forward prop

devrnt:
### Info
Add `shouldForwardProp` to hide to boilerplate for the `forwardProp` function in `setup`.
The user should pass a filter function that resolves to a boolean. Truthy results of the filter function will be let through,  falsy results will be removed from the `props`.

See https://github.com/cristianbote/goober/pull/227#discussion_r553164138

### Use
```jsx
import * as React from 'react';
import { setup } from 'goober';
import { shouldForwardProp } from 'goober/forwardProp';

setup(
  React.createElement,
  undefined,
  undefined,
  shouldForwardProp((prop) => {
    return prop[0] !== '$';
  }),
);
```

~~Don't mind commit `7bf7ff9` it's in sequence of https://github.com/cristianbote/goober/pull/233~~ 
I did a rebase


cristianbote:
woow! 😄 this looks great!

package.json
```diff
@@ -49,6 +53,12 @@
       "import": "./prefixer/dist/goober-autoprefixer.mjs",
       "require": "./prefixer/dist/goober-autoprefixer.js",
       "umd": "./prefixer/dist/goober-autoprefixer.umd.js"
+    },
+    "./forwardProp": {
```
cristianbote, line 57: How do you feel if we rename the `forwardProp` in `shouldForwardProp`? 🤔 it's longer, but might look more familiar for folks coming from other libraries.

devrnt: Yeah seems good to me!

yarn.lock
```diff
@@ -0,0 +1,6752 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
devrnt, line 1: Will drop this `yarn.lock`

package.json
devrnt: Do you want these reference to be changed everywhere (like in `forwardProp/package.json`) or just this "directory export"?

cristianbote: How about everywhere for the sake of consistency? So the name should be camelCase (`shouldForwardProp`) and the folder should be dash-separated(`should-forward-prop`).

forwardProp/package.json
```diff
@@ -0,0 +1,40 @@
...
+  "scripts": {
+    "build:lib": "rm -rf dist && microbundle --entry src/index.js --name gooberForwardProp --no-sourcemap",
+    "test": "jest --setupFiles ./jest.setup.js"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/cristianbote/goober.git",
+    "directory": "forwardProp"
+  },
+  "author": "Cristian <botecristian@yahoo.com>",
```
cristianbote, line 22: While this indeed is under my user's scope, this package is all on you @devrnt! 😃  So, I would encourage you to stamp your name in here! 🎉

devrnt: Okay changes noted 😀

package.json
devrnt: Sounds good to me!

shouldForwardProp/should-forward-prop.d.ts
```diff
@@ -0,0 +1,9 @@
+export = gooberGlobal;
```
cristianbote, line 1: Nit: I think this should be changed to 
```suggestion
export = gooberShouldForwardProp;
```

right?

shouldForwardProp/should-forward-prop.d.ts
```diff
@@ -0,0 +1,9 @@
...
+
+export as namespace shouldForwardProp;
+
+declare namespace gooberGlobal {
```
cristianbote, line 5:
```suggestion
declare namespace gooberShouldForwardProp {
```

package.json
```diff
@@ -20,6 +20,10 @@
     "prefixer/package.json",
     "prefixer/autoprefixer.d.ts",
     "prefixer/README.md",
+    "shouldForwardProp/src",
```
cristianbote, line 23: Uh needs the `shouldForwardProp/dist` as well. But I'll take care of it later 👍

cristianbote (approved):
woop woop! 🙌  awesome!

shouldForwardProp/src/index.js
```diff
@@ -0,0 +1,9 @@
+export function shouldForwardProp(filterPropFunction) {
+    return function forwardProp(props) {
+        for (let p in props) {
```
devrnt, line 3: Will call this `prop` for convenience

cristianbote (approved):

PR #236, opened 2021-01-15
feat: add clx utility for combining classes 

### LINKED ISSUE #185: Handling specificity when using the css function.
Hi there! 👋

I've been using the `css` function to handle styles but ran into an issue of specificity when trying to combine them. [Here's an example sandbox](https://codesandbox.io/s/competent-star-62syf?file=/src/App.js).

The issue here seems to be that goober doesn't provide a mechanism to combine class names while overriding them in the correct order. Emotion supports this through their own helper ([linked here](https://emotion.sh/docs/emotion#cx)). I'm not sure if this is something that goober is able to (or wants to) support, but I figured I'd mention it 😅

I did see #98 which looked similar, but it seemed like the discussion was about using the `styled` function instead of `css` which can require a different component architecture. Ideally, a user would be able to combine generated class names such that they override styles based on the order they are specified.

---

Also wanted to mention that this is a great library that I'm happily using in a few of my projects. Thank you 🙏

jakehamilton:
This PR adds a utility function named "clx" to the base "goober"
module. The "clx" function takes one or more class names and creates
new scoped styles based on the order the class names were supplied.

resolves #185


src/clx.js
```diff
@@ -0,0 +1,35 @@
...
+import { toHash } from './core/to-hash';
+import { getSheet } from './core/get-sheet';
+
+/**
+ * clx
+ * @param {...String} classes
+ */
+function clx() {
+    let ctx = this || {};
+    let classes = [].reduce.call(arguments, (classes, c) => classes.concat(c.split(' ')), []);
```
jakehamilton, line 12: I tried to keep the coding style the same as other files. Typically I would've used `const`, rest/spread, and others but I wasn't sure about the compilation configuration and/or coding conventions.

src/clx.js
```diff
@@ -0,0 +1,35 @@
+import { cache } from './core/hash';
```
jakehamilton, line 1: If we want to keep `clx` out of the core, this cache will need to be exported from `goober` itself. Currently it's only available from the `core/hash` module directly.

README.md
```diff
@@ -277,7 +278,7 @@ setup(React.createElement, undefined, undefined, (props) => {
 });
 ```
 
-The functionality of "transient props" (with a "$" prefix) can be implemented as follows:
+The functionality of "transient props" (with a "\$" prefix) can be implemented as follows:
```
jakehamilton, line 281: It looks like this got hit with the formatter. The rendered result is unchanged👍

src/clx.js
```diff
@@ -0,0 +1,35 @@
...
+ * clx
+ * @param {...String} classes
+ */
+function clx() {
+    let ctx = this || {};
+    let classes = [].reduce.call(arguments, (classes, c) => classes.concat(c.split(' ')), []);
+
+    let id = toHash(classes.join(''));
+
+    let hashedClasses = classes.reduce((hashedClasses, c) => {
```
jakehamilton, line 16: This used to be structured differently, but i dont think this needs to be a reduce anymore. Map will be smaller and more efficient 💯

cristianbote:
@jakehamilton this is really awesome! Thank you for taking the time to open this PR! Please allow me a few days to go through it and understand it properly. Been a bit busy lately and couldn't focus on this.

jakehamilton:
@cristianbote no worries! I'm in no rush here, so please take all the time you need. I'm also happy to answer any questions you may have. Cheers!

README.md
cristianbote: It's totally fine 😃

jakehamilton:
Hey @cristianbote! Wanted to circle back around and see if you've had time to think about this feature.

cristianbote:
Hi @jakehamilton! Each time I open the PR tabs I am thinking about it 😄 . Honestly the most viable option, size-wise, would be to have a separate package for it. So that would mean:

```js
import { css } from 'goober';
import { clsx } from 'goober/clsx';
```

This will still have a size increase, because the _cache_ would need to be exported from the main package. 😬  But less than having the `clsx` part of the main package.

So, I suppose that means a separate package? 🤔

jakehamilton:
Hey @cristianbote!

I think I've also landed on making this a separate package. It's a small feature so it should be easy to port 👍

How do you feel about an initial PR that only exports the `cache`? That way we could test things out before making an official `goober/clsx` package. Alternatively, if you want I can draft a PR that exports `cache` _and_ creates a new `goober/clsx` package using my current changes. Any particular preference?

I know this is an important thing to get right so I don't want to rush it!

---

Yak Shaving:

For naming of the helper and its package, were you thinking `clsx`? I'm not opposed to it, but have found `clx` is nice and quick to type 😉

jackyef:
Hey 👋 , just wanted to say that I am encountering the same issue currently and having this feature would be really great!

jakehamilton:
Hey @jackyef 👋 

I've been a bit distracted with other work, so not sure if/when I'll get back around to this issue. As I left off with @cristianbote, the discussion was about separating this `clsx` feature into a new package so as not to bloat the core. If the implementation here is acceptable then all that needs done is separating the `clx.js` file into its own sub-package (similar to others in the repository).

cristianbote:
Hey folks!

I believe thanks to the The Great Shave Off Challenge this is pretty doable now, size-wise? As far as I remember that was the only concern really. So, should we? 😄

jakehamilton:
Heya!

I think since there will probably be a good amount of folks who'll only be using the `styled` variant, it might still be a good idea to separate this functionality. Making it opt-in sounds like a good idea for me.

jakehamilton:
Closing in favor of #350.

Issue #209, opened 2020-12-08
css prop functionality as part of babel-plugin and macro


cristianbote:
Examples:
- emotion https://emotion.sh/docs/css-prop
- styled-components https://styled-components.com/docs/api#css-prop

cristianbote:
Jus discovered this https://github.com/agneym/babel-plugin-goober-css-prop 😲 awesome job @agneym! I'll add it to the list of integrations if you don't mind.

Thank you!

agneym:
Sure 😄 I made this so I could try out `tw` prop offered by [twin.macro](https://github.com/ben-rogerson/twin.macro). 

[sample repo](/agneym/next-goober-twin)

I haven't tried out other use cases for it though, hopefully it works for others 😅

cristianbote:
I might be able to look more into it and test it, but meanwhile feel free to open a PR to add it in here https://github.com/cristianbote/goober#integrations ✌️

PR #237, opened 2021-01-17
add babel-plugin-css-prop to integrations

agneym:
Relates to #209 

* Adds [babel-plugin-goober-css-prop](https://github.com/agneym/babel-plugin-goober-css-prop) to the integrations list.
* Corrects link for [babel-plugin-macros](https://github.com/kentcdodds/babel-plugin-macros)

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,8 @@ It's a pun on the tagline.
     -   [Babel Plugin](#babel-plugin)
     -   [Babel Macro Plugin](#babel-macro-plugin)
     -   [Gatsby](#gatsby)
+    -   [Preact CLI Plugin](#preact-cli-plugin)
+    -   [CSS Prop](#css-prop)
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -534,6 +534,34 @@ export default (config, env) => {
 
 When you build your Preact application this will run `extractCss` on your prerendered pages and add critical styles for each page.
 
+## CSS Prop
```
cristianbote, line 537: Would you mind adding this as part of TOC at the top as well? Also, I think this would be sweet to be included in the generated docs at goober.js.org. So that means, this section needs to be copy/pasted into `/docs` folder in the hierarchy.

agneym: Updated both changes

cristianbote (approved):
This is freaking great @agneym! Thank you!

PR #238, opened 2021-01-18
(feat) Increase the SSR perf

cristianbote:
This change increases the SSR perf at least 3x while keeping the browser env consistent.


cristianbote:
For reference https://github.com/cristianbote/goober/pull/238/checks?check_run_id=1723876595#step:4:168

mauadi (approved):

PR #239, opened 2021-01-18
Add false as possible type in styled props

devrnt:
### Problem
```tsx
const Box = styled('div')<{ border: boolean }>`
   color: red; 

   ${({ border }) =>
       border && `
         border: 1px solid red;
      `
  }
`;
```

#### TSC output: 
```
...
Type 'false' is not assignable to type 'string | number | CSSAttribute | undefined'.
```

### Current workaround 
Use ternary operator and return empty string for the falsy value

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html


src/__tests__/styled.test.js
```diff
@@ -146,4 +146,14 @@ describe('styled', () => {
 
         expect(vnode).toMatchVNode('tag', { className: 'go', bar: 1 });
     });
+
+    it('pass truthy logical and operator', () => {
```
devrnt, line 150: Actually can be dropped since it's kind of verbose. I just added this here to verify the `&&` operator

cristianbote: But this test does not run with TypeScript 🤔  does it surface the same problem in JavaScript as well?

devrnt: No, it's working like expected. In general this "issue" can be solved by just ignoring the tsc on the user's end (aka adding `@ts-ignore`)

cristianbote: Oh, I see! 😄  Do you think we can capture it in here instead? https://github.com/cristianbote/goober/blob/master/ts-tests/test-api.tsx

devrnt: Okay I added them there. Undoing the `global.d.ts` update of this PR will make the TSC "complain" for the `<BoxWithLogicalAnd />`

cristianbote (approved):
👍  looks great!

cristianbote:
Published `2.0.25` with this fix

PR #240, opened 2021-01-19
fix should forward prop types reference

devrnt:
The `types` field in `package.json `for `should-forward-prop`  is pointing to an invalid type file


cristianbote (approved):

cristianbote:
Published `2.0.26`

PR #250, opened 2021-02-04
chore: support goober prefixer in gatsby plugin

devrnt:
Support both `goober-autoprefixer` and `goober/prefixer` in `gatsby-plugin-goober`

### DIFF
--- a/packages/gatsby-plugin-goober/gatsby-ssr.js
+++ b/packages/gatsby-plugin-goober/gatsby-ssr.js
@@ -1,9 +1,22 @@
 const React = require('react');
 const goober = require('goober');
-const autoprefixer = require('goober-autoprefixer');
+
+const moduleExists = (path) => {
+    try {
+        require.resolve(path);
+        return true;
+    } catch (_) {
+        return false;
+    }
+};
+
+// Support new and old goober versions
+const autoprefixer = moduleExists('goober/prefixer')
+    ? require('goober/prefixer').prefix
+    : require('goober-autoprefixer').default;
 
 // Set the pragma for `goober` to know which to use
-goober.setup(React.createElement, autoprefixer.prefix);
+goober.setup(React.createElement, autoprefixer);
 
 // Define a cache entry
 const cache = new Map();


packages/gatsby-plugin-goober/gatsby-ssr.js
```diff
@@ -1,9 +1,22 @@
...
+};
+
+// Support new and old goober versions
+const autoprefixer = moduleExists('goober/prefixer')
+    ? require('goober/prefixer').prefix
+    : require('goober-autoprefixer').default;
 
 // Set the pragma for `goober` to know which to use
-goober.setup(React.createElement, autoprefixer.prefix);
+goober.setup(React.createElement, autoprefixer);
```
devrnt, line 19: Since we don't pass any theme or filter here function, does that mean this functionality can't be used?

cristianbote: Hey @devrnt sorry, been away for a while. That's a good point. 🤔  No idea. At first glance I think you are right, wondering if this is something that the community is using though? 

Recently I believe the ecosystem is moving towards using css vars instead of theme with context.

cristianbote (approved):

packages/gatsby-plugin-goober/gatsby-ssr.js
devrnt: No rush @cristianbote 😀. Yes indeed, but I still believe that a lot of people are looking for such a "global theme pragma" since they are migrating from another css-in-js.

Anyway it's out of the question here, I just came across this while I was looking at the gatsby starter

cristianbote:
@devrnt about to publish a new version `2.3.0` but do you think we could drop the `goober-autoprefixer` and use the addon instead `goober/prefixer`?

devrnt:
@cristianbote As far I can see these are the same: https://github.com/cristianbote/goober/blob/master/package.json#L53 so you should be able to drop it and use the add-on.

cristianbote:
Ok. 😄  Gonna do the changes and publish a new patch! The addon has changed since then -- just last night I've switched to `style-vendorizer` to fix some discrepancies.

devrnt:
Hadn't see these changes, looks goed! 😀

PR #255, opened 2021-02-27
Add Typings for Arrays of CSS Attributes

kerimhudson:
When utilising TypeScript with Goober, and the array method of adding CSS attributes, like so:

```ts
import { styled } from 'goober'

styled('div')([{
  color: 'blue'
}])

```
TypeScript complains that an array cannot be used, however this is valid Goober syntax (see screenshot below).
Screenshot 2021-02-27 at 15 46 53

This PR adds the option for an array to passed as `CSSAttribute[]` to stop TypeScript complaining.

### DIFF
--- a/goober.d.ts
+++ b/goober.d.ts
@@ -67,6 +67,7 @@ declare namespace goober {
     type Tagged<P extends Object = {}> = <PP extends Object = {}>(
         tag:
             | CSSAttribute
+            | CSSAttribute[]
             | TemplateStringsArray
             | string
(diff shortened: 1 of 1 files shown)


cristianbote:
Thank you @kerimhudson! 🙏

kerimhudson:
> Thank you @kerimhudson! 🙏

No worries. It's not perfect yet, as there's still some edge cases I'm noticing that aren't working, but I'll get around to them when I have some time.

PR #262, opened 2021-03-18
Give preference to require main

guybedford:
This came up in https://github.com/jspm/project/issues/96 on jspm.dev for react-host-toast.

react-hot-toast contains:

```js
goober = require('goober');
```

When JSPM builds for the browser, it converts the above into:

```js
import goober from 'goober';
```

Per the Node.js semantics. But then uses the browser ESM version, which doesn't export a default export.

This ensures the require version is always used for CJS requires both in the browser and Node.js. An alternative could be to provide a default export in the ESM version as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,16 +34,12 @@
   ],
   "exports": {
     ".": {
+      "require": "./dist/goober.js",
       "browser": "./dist/goober.modern.js",
       "import": "./dist/goober.esm.js",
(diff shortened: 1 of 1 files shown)


cristianbote:
Hey @guybedford! Thank you so much for looking into this and fixing it up. Will patch it up quickly.

Issue #257, opened 2021-03-08
Function components cannot be given refs.


Crisis2010:
Hello! I'd like to move from styled-components to goober, but can't fix one issue:

`Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?`

Here is example:
https://codesandbox.io/s/react-goober-forked-w3dh2?file=/src/index.js

i've tried many things, but `console.log(ref)` is always undefined, instead of el
What i'm doing wrong?

cristianbote:
Hi @Crisis2010, thanks for opening this issue.

Since `goober` is agnostic of your virtual dom, jsx, library of choice,`styled` supports a second argument that is the reference for your library `forwardRef` API. In your case you need to pass `React.forwardRef` as the second argument to the components that you need your ref to:

```jsx
const Title = styled("h1", React.forwardRef)`
  font-weight: bold;
  color: dodgerblue;
`;
```

Let me know if this helps. Cheers!

Crisis2010:
Yeah, thanks! Tried everything except this)

It seems to me that it will not be superfluous to indicate this more explicitly in the documentation.

cristianbote:
Hey @Crisis2010, oh you are right! It should totally be reflected in the docs. Right now it only appears only as part of the documentation of `styled` function https://github.com/cristianbote/goober#styledtagname-string--function-forwardref-function I believe another example in that section would help. Would you like to add it? 😄 I would be really grateful! 🙏

mpkelly:
@cristianbote I needed this too and didn't spot it in the docs. So I have added an example to make it more obvious and sent you a PR. It would be handy to keep this param of `styled` but allow a default as part of the `setup` call. I could send you a PR later for that, if you want?

Also have a question: does goober have an equivalent of [.attrs](https://styled-components.com/docs/api#attrs) from styled-components? I am looking for a way to add/remove props that the component receives.

cristianbote:
Awesome work @mpkelly! Thanks for opening the PR.

As for the `.attrs` api, there's no equivalent in goober. I believe there should be a hard line between _Styled_ components and regular ones and I feel strange about `.attrs` 😄  it's like a _pre_-processor for your props but shouldn't that functionality exist inside the functional components instead? I don't know, I really think we should keep these things separated.

Anyway in goober the equivalent would be to have a `InputStyled` that handles the styling and the main `Input` component handling the type and size. Like so:

```jsx
import { styled } from "goober";

[...]

PR #268, opened 2021-03-22
Add forward ref function example to API docs

mpkelly:
Based on CB's request here

https://github.com/cristianbote/goober/issues/257


cristianbote:
Huge thanks @mpkelly! This is great! Do you mind adding the exact same thing in here as well? https://github.com/cristianbote/goober/blob/master/docs/docs/api/styled.md

cristianbote:
woop woop! Lighting fast @mpkelly! 😄

PR #276, opened 2021-04-02
fix: Missing TS types for `styled` function (#275)

### LINKED ISSUE #275: Missing types for `styled` with array of functions
## Problem

The documentation has an example that allows passing an array to the `styled` function whose items can also be a function  (https://goober.js.org/api/styled#arrays). Unfortunately the current TS typings do only allow an array of `CSSAttribute` (https://github.com/cristianbote/goober/blob/master/goober.d.ts#L70).

## Example

```ts
type ButtonProps = {
  isLoading?: boolean
}

styled("button")<ButtonProps>([
  { color: "red" },
  ({ isLoading }) => ({ color: "blue" }) // 👈 TS complains about `({ isLoading }) => ({ color: "blue" })` is not assignable to `CSSAttribute`
])
```

## Fix

A simple solution would be to just extend the existing `CSSAttribute[]` [here](https://github.com/cristianbote/goober/blob/master/goober.d.ts#L70) to also accept a function:

```ts
type StylesGenerator<P extends Object = {}> = (props: P) => CSSAttribute | string
type Tagged<P extends Object = {}> = <PP extends Object = {}>(
        tag:
            | CSSAttribute
            | (CSSAttribute | StylesGenerator<P & PP>)[]
            | TemplateStringsArray
            | string
            | StylesGenerator<P & PP>,
        ...props: Array<
            | string
            | number
            | ((props: P & PP) => CSSAttribute | string | number | false | undefined)
        >
    ) => StyledVNode<Omit<P & PP, keyof Theme<DefaultTheme>>>;
```

If you want i would be happy to open a PR!

steschwa:
- Fix #275 
- Add test to ensure functions are valid as array items for `css(...)`

### DIFF
--- a/goober.d.ts
+++ b/goober.d.ts
@@ -64,13 +64,14 @@ declare namespace goober {
         ...props: Array<string | number>
     ): string;
     type StyledVNode<T> = (props: T, ...args: any[]) => any;
+    type StylesGenerator<P extends Object = {}> = (props: P) => CSSAttribute | string;
     type Tagged<P extends Object = {}> = <PP extends Object = {}>(
         tag:
             | CSSAttribute
-            | CSSAttribute[]
+            | (CSSAttribute | StylesGenerator<P & PP>)[]
             | TemplateStringsArray
             | string
-            | ((props: P & PP) => CSSAttribute | string),
+            | StylesGenerator<P & PP>,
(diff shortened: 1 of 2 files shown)


cristianbote (approved):
🙏  thank you so much!

Issue #279, opened 2021-04-08
Should I use goober setup once or on every component?


pure-js:
I want to clarify whether you need to use `setup` on each page or once for the entire SPA?
```javascript
import React from 'react';
import { setup } from 'goober';

setup(React.createElement);
```

cristianbote:
Hey @pure-js, only once at the root of your app. Gonna update the docs and readme to reflect this as well. Thanks for pointing it up! 😃 If you want feel free to open the PR yourself. 👍

PR #283, opened 2021-04-12
Updates docs regarding setup() - once only - in entry file.

mburridge:
Adds the following line to the docs on `setup()`:

The call to `setup()` should occur once only. It should be called in the entry file of you project.


cristianbote:
Fixes #279 as well. Awesome work @mburridge! Thank you!

PR #271, opened 2021-03-24
Minor rewrite of hashing function to improve performance by ~3x

danielweck:
Goober's current hashing function causes unnecessary memory allocations (`Array.split()`) as well as compute cycles / function calls (`Array.reduce()`). The following simple change does not increase the bundle size, and yields much better performance (at least 2x faster, nearly 3x):

```diff
-export let toHash = (str) =>
-    'go' + str.split('').reduce((out, i) => (101 * out + i.charCodeAt(0)) >>> 0, 11);
+export let toHash = (str) => {
+    let i = 0, l = str.length, out = 11;
+    while (i < l) out = (101 * out + str.charCodeAt(i++)) >>> 0
+    return 'go' + out;
+}
```

Comparisons using Benchmark.js:

`npm run test-perf-hash`

=>

```
> NODE_ENV=production node benchmarks/perf-hash.js


Starting HASH!
▸ goober original HASH x 3,180,521 ops/sec ±1.06% (87 runs sampled)
▸ goober optimized HASH x 8,219,240 ops/sec ±1.36% (84 runs sampled)

Fastest is: goober optimized HASH
```

```
> NODE_ENV=production node benchmarks/perf-hash.js


Starting HASH!
▸ goober original HASH x 3,582,669 ops/sec ±0.92% (90 runs sampled)
▸ goober optimized HASH x 9,067,875 ops/sec ±1.42% (94 runs sampled)

Fastest is: goober optimized HASH
```

```
> NODE_ENV=production node benchmarks/perf-hash.js

Starting HASH!
▸ goober original HASH x 3,517,773 ops/sec ±1.02% (89 runs sampled)
▸ goober optimized HASH x 10,189,264 ops/sec ±1.81% (90 runs sampled)

Fastest is: goober optimized HASH
```

```
> NODE_ENV=production node benchmarks/perf-hash.js

Starting HASH!
▸ goober original HASH x 3,368,369 ops/sec ±1.45% (85 runs sampled)
▸ goober optimized HASH x 9,699,551 ops/sec ±1.49% (90 runs sampled)

Fastest is: goober optimized HASH
```

PR #288, opened 2021-04-15
Golf some low hanging bytes

kitten:
Ok, this is mostly some low hanging fruit, but can't leave the really easy stuff over, now that @JoviDeCroock told me about the "Great Shave Off" :D 

One thing I'm noting is that there were two `str[0]` checks where afaict `str` suffices, but did those serve any purpose before, i.e. am I missing something?

Locally this produces:

```
1173 B: goober.js.gz => 1161 B: goober.js.gz (-12B)
1181 B: goober.modern.js.gz => 1169 B: goober.modern.js.gz (-12B)
```

Edit: by the way, it's possible to get to `goober.js.gz 1156B` quite easily by refactoring `astish` to identity return objects, and always hash (instead of using the cache for the hashes too). This does run `hash` more often and makes some arguments more polymorphic, however, if the hash function is turned into something that JS engines like more (specifically djbv2) the perf regression can be minimised and you can reach `1155B` i.e. an additional `-6B` :tada:

I haven't pushed that here since I'd have to update some tests but this is the commit with an explanation: https://github.com/kitten/goober/commit/eafd29d2588e6dc02121c17d8a8bd8546049d37c

Edit 2: Ok, this may not save that much unless you handle `go-` cases since it'd have to drop the unsigned int32 hint to still save more bytes. Bringing that back it's `1159B`


cristianbote:
@kitten this is amazing! 🎉  😄  So happy you decide to chip in!

These are amazing finds! Gonna look into that `str[0]` cause I fail to remember what it was for.

src/core/hash.js
```diff
@@ -14,14 +14,13 @@ let cache = {};
...
-    let out = '';
-
-    for (let p in data) {
-        let val = data[p];
-        out += p + (typeof val == 'object' ? stringify(data[p]) : data[p]);
+    if (typeof data == 'object') {
+        let out = '';
+        for (let p in data) out += p + stringify(data[p]);
+        return out;
+    } else {
```
danielweck, line 21: no need for else here? (return statement in if branch)

src/core/astish.js
```diff
@@ -12,11 +12,11 @@ export let astish = (val) => {
 
     while ((block = newRule.exec(val.replace(ruleClean, '')))) {
         // Remove the current entry
-        if (block[4]) tree.shift();
-
-        if (block[3]) {
+        if (block[4]) {
+            tree.shift();
+        } else if (block[3]) {
```
danielweck, line 17: I imagine you have thought this through, but just in case the unit tests don’t cover this edge case: this code change modifies the logic when block[4] AND block[3] are truthy. Is that an oversight, or intentional?

danielweck:
Regarding the hash function, note that cyrb32 with Math.imul() is quite performant. See https://github.com/cristianbote/goober/pull/271

kitten:
@danielweck totally agreed, and I did consider it since the same goes for djbv2 and any other similar variant 👍 I didn't drop it in since I'd assume it'd increase the size which isn't this project's goal right now (?)

As far as I'm concerned it's browser CSS engine characteristics first, string operations second, and micro-optimisations last in what matters. So while it'd yield a performance uplift in synthetics, as long as performance currently doesn't drop significantly, I'd wager that Goober is "fast enough" if compared to the browser's overhead with styling.

src/core/astish.js
kitten: That's intentional; this can't happen. I'm not sure if it was possible in the past but a closing curly is in a separate RegEx alternation

src/core/hash.js
kitten: I did compare this. That's just a coding style thing, and I was focusing on what ultimately is output (which is likely nested ternaries on a single statement)

Happy to change this depending on what this repo prefers given that Terser will likely output the same code for either 👍

danielweck: thank you :)

src/core/astish.js
danielweck: great :)

cristianbote: 🙌  hehe 😄  nice!

src/core/hash.js
cristianbote: Honestly if prettier didn't complain or change it 😄  then it's totally fine by me! Good find by the way!

danielweck: https://eslint.org/docs/rules/no-else-return

;)

src/styled.js
```diff
@@ -59,7 +59,7 @@ function styled(tag, forwardRef) {
             }
 
             // Handle the forward props filter if defined and _as is a string
-            if (fwdProp && _as[0]) {
+            if (fwdProp && _as) {
```
cristianbote, line 62: Hey @kitten, so this is used in here to flag that only at the DOM tag level we need to run the fwdProp function in order to remove them from appearing onto the element as attributes. So this essentially is a poor man version of `typeof _as === "string"`.
```suggestion
            if (fwdProp && _as[0]) {
```

src/core/to-hash.js
```diff
@@ -8,8 +8,9 @@
  * @returns {String}
  */
 export let toHash = (str) => {
-    let i = 0,
+    let char,
+        i = 0,
         out = 11;
-    while (i < str.length) out = (101 * out + str.charCodeAt(i++)) >>> 0;
+    while ((char = str.charCodeAt(i++)) || char == char) out = (101 * out + char) >>> 0;
     return 'go' + out;
```
cristianbote, line 15: With these changes the overall size increases 🤔 with `+5B`. Is there something that I am not checking correctly?

cristianbote:
@kitten THESE ARE AMAZINGLY great finds! With the above changes the total amount of bytes shaved is <drumrolls> `-10B` boom! 🎉 Well done! Don't forget to submit your hard earn invoice on opencollective! 😉 

Thank you for contributing!

src/core/to-hash.js
```diff
@@ -8,8 +8,9 @@
  * @returns {String}
  */
 export let toHash = (str) => {
-    let i = 0,
+    let char,
+        i = 0,
```
cristianbote, line 12:
```suggestion
    let i = 0,
```

src/core/to-hash.js
```diff
@@ -8,8 +8,9 @@
...
         out = 11;
-    while (i < str.length) out = (101 * out + str.charCodeAt(i++)) >>> 0;
+    while ((char = str.charCodeAt(i++)) || char == char) out = (101 * out + char) >>> 0;
```
cristianbote, line 14:
```suggestion
    while (i < str.length) out = (101 * out + str.charCodeAt(i++)) >>> 0;
```

PR #291, opened 2021-04-15
Small grammar and typo fixes in README

jevakallio:
👋 Hey! I came over from Twitter via The Great Shave Off. 

Didn't end up shaving any bytes, but I did polish the README a little bit. I tried to make sure not to change any semantics or tone, just grammar and typos.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,11 +17,11 @@
 
 # 🪒 The Great Shave Off Challenge
 
-Can you shave off bytes from goober? Do it and your gonna get paid! [More info here](https://goober.rocks/the-great-shave-off)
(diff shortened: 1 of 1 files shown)


cristianbote:
@jevakallio this is... amazing! 😲  Thank you so much for this! 🙏  I don't know how you do it, but this is top notch! 🙇 

Don't worry about the bytes 😄  in my book this is as important as that. 😉  It's rare when someone takes form their free time to put that much into correcting other folks typos.

We also have a `/docs` folder where we basically keep each README section to generate the documenation website out of them. Not super happy about the flow, but it is what it is. I'm planing to generate the docs out of the main README so the inout is the same.

PR #298, opened 2021-04-22
Shavings

rpearce:
Hi! I saw https://goober.rocks/the-great-shave-off and thought I'd take a few minutes and see what I could shave off.

I ran the build command locally and compared before and after change sizes, so there should be a few here!

Let's see what your github action says... this may be garbage.

### DIFF
--- a/src/core/compile.js
+++ b/src/core/compile.js
@@ -10,12 +10,12 @@ export let compile = (str, defs, data) => {
         let tail = defs[i];
 
         // If this is a function we need to:
-        if (tail && tail.call) {
+        if ((tail || {}).call) {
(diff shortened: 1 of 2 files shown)


rpearce:
Yup, this was garbage. Closing.

cristianbote:
Hey @rpearce! Yeah 😅  gzip it's really really favorable to repeatable patterns. You can get visual feedback of the compressions here as well https://gzthermal.now.sh/. I use it regularly to get a sense of what can be a bit more compressed.

Don't give up! 😃  there's definitely some bytes there that need shaving!

rpearce:
Thanks for the inspiration!

PR #301, opened 2021-05-02
shrink style id prefix

alexanderchan:
Are we up for shrinking down the id a bit?  We probably don't risk collisions too much by shrinking to `_gbr` and still can have an identifier that is readable.  We could go even smaller (`_gb` or `_g`) but it it worth sacrificing readability for size?

### DIFF
--- a/src/core/get-sheet.js
+++ b/src/core/get-sheet.js
@@ -1,4 +1,4 @@
-let GOOBER_ID = '_goober';
+let GOOBER_ID = '_gbr';
 let ssr = {
     data: ''
 };


cristianbote:
Hey @alexanderchan thank you so much for your submission!

Took a look at the results table:

**Size Change:** -4 B (0%) 

**Total Size:** 4.74 kB

| Filename | Size | Change | |
|:--- |:---:|:---:|:---:|
| `dist/goober.esm.js` | 1.17 kB | -2 B (0%) |  |
| `dist/goober.js` | 1.16 kB | -2 B (0%) |  |
| `dist/goober.modern.js` | 1.17 kB | -2 B (0%) |  |
| `dist/goober.umd.js` | 1.24 kB | +2 B (0%) |  |

It does shave `-2B` which is awesome! But I would rather keep the existing `_goober` value as an ID. There is an ecosystem around goober which rely on predefining the `style` tag from the server or statically and they use `_goober` as their ID value.

alexanderchan:
@cristianbote totally understand!  thanks for the awesome library, it's been a great pleasure to read through the code.  I wasn't aware of the ecosystem part and that's good to know about.  Appreciate your work on all of this.  I'll close this.

PR #313, opened 2021-05-15
Save 1b in css.js

dios-david:
Hey,

I recently heard about The Great Shave Off, which I think is a really fun idea! 🥳 

I spent some time today trying to save some bytes here and there, and I had some great results but after the gzip compression usually the size got bigger 🤔 
This is the only thing I could find (yet), and I'm not 100% convinced this is a right thing to do (calling `Object.assign` even when `i` is falsy), but it saves a byte so thought I submit it for review.


cristianbote:
Hey @dios-david!

Thank you so much for submitting this!

That's the gzip _dance_ 😄  The more you use a syntax the better is the compression. At the beginning I was able to add full features with `0B` due to gzip.

I have to check the perf bench to make sure there's no degradation. I'll get back to you.

cristianbote:
Hey @dios-david, thank you so much for opening this PR!

After checking the perf -- which I am sorry that it took this much, there's a slight tiny really degradation for conditional array components:
```js
const Foo = styled('div')(props => [
  { /* base styles */ },
  props.baz && {
    baz: 1
  }
])
```
but we can live with it! It's under `1ms` and I have plans to improve perf in other areas which will offset this for sure. So with this being said, let's merge this in! 🙌 Congrats for the find @dios-david! Don't forget to submit for your hard earn buck!

PR #324, opened 2021-06-14
Fix: add all valid chars to css rule regexp

gr0uch:
Problem: valid css declarations may include the underscore (`_`) char. I believe this was just an oversight or typo.

Example:
```
:root {
  --valid_name: 0;
}
```

Before this fix, goober would mangle this:
```
:root {
  --valid: _name: 0;
}
```

After fix:
```
:root {
  --valid_name: 0;
}
```
as expected.


cristianbote:
That's a good catch @gr0uch! Definitely an oversight. Would you mind adding a test case for this as well? That would make this really future proof.

gr0uch:
@cristianbote alright, i added a test case for this.

src/core/astish.js
```diff
@@ -1,4 +1,4 @@
-let newRule = /(?:([A-Z0-9-%@]+) *:? *([^{;]+?);|([^;}{]*?) *{)|(})/gi;
+let newRule = /(?:([\u0080-\uFFFFa-zA-Z0-9-_%@]+) *:? *([^{;]+?);|([^;}{]*?) *{)|(})/g;
```
gr0uch, line 1: btw i turned off case-insensitivity here: https://stackoverflow.com/questions/32010/is-regex-case-insensitivity-slower

cristianbote: Uh oh! 😬  this adds `+15B` 😲

gr0uch: the negative set might be even longer tbh, because you'd have to exclude alphanumerics and the `-_%@` chars and the same unicode range, can't really think of a way to make it shorter. i'd rather have a lib that works correctly according to specs, because 15bytes isn't going to make a difference in the real world, not even if we were still using dialup modems.

gr0uch: this is now +8 bytes, due to writing out `\u0080-\uFFFF`. it can be omitted to even shave some bytes but it wouldn't be correct.

cristianbote:
@gr0uch this is amazing! 👍  nicely done! You are right, it's a small price size wise to be correct. Thank you so much for taking care of this 💯

PR #325, opened 2021-06-15
fix: Add a target argument to the extractCss typing

sunnyone:
extractCss() can receive a target argument, but d.ts does not support it.
https://github.com/cristianbote/goober#extractcsstarget


cristianbote:
Thank you!

PR #339, opened 2021-07-08
export useTheme hook

jlalmes:
Missing functionality that exists in `styled-components`


cristianbote:
Hey @jlalmes! Thank you for opening this one.

So, I have to check the size increase on exporting this named function `useTheme`. Why it doesn't existed before is that usually the `useTheme` is provided by the user in the `setup()` call. So it didn't made a lot of sense to have it re-exported.

Basically this is what's happening:

```js
// user theme code
const themeContext = createContext({ /* this is where the theme props are going */ });

const useTheme = () => {
  return useContext(themeContext)
}

// Then the `setup` call is happening
setup(
  React.createElement,
  undefined,
  useTheme // <-- this is where the user's `useTheme` hook is set.
)
```

So that would mean `goober` will re-export a user provided function. If this would not have a size impact I don't think it would be an issue, but having a named export called `useTheme` is `+8B`. What are your thoughts on this? Is the size increase worth it overall?

jlalmes:
Hi @cristianbote - thanks for checking this out! Let me explain our use case. 

We are using a monorepo pattern where each React app will `setup` goober with its custom theme. Then in our generic shared component library we can call `useTheme` to get access to the theme setup by the parent app. So although goober will just be "re-exporting" the user-provided theme function - our components are not currently able to import it in any other way. 

Our project structure:
```
- apps
  - auth (calls `goober.setup`)
  - dashboard (calls `goober.setup`)
  - js-sdk (calls `goober.setup`)
- libs
  - components (uses `useTheme`)
```

I think we could rework the components and theme into a global context object, but exporting `useTheme` seems like a no-brainer solution. I understand you are concerned about the `+8B` though and I guess the extreme bare-bones nature of this project is it's appeal. I won't be offended if you reject this 😅.

PR #350, opened 2021-07-26
feat: add clsx utility as a subpackage

jakehamilton:
> **NOTE**: These changes are quite similar to #236, but I've gone ahead and made a new PR based off of the latest changes on `master`.

Well, here we are again! I found some time to work on this `clsx` utility to let us combine & override classes. The current implementation is very similar to what I had in #236, but there are a few important notes here:

1. Since I had separated `clsx` into its own subpackage, it now depends on `goober` exposing certain core utilities.
2. With my current changes, `goober` now exposes an object `core` which contains the necessary functions for `clsx` to work.
3. I've tried to keep the API change minimal so as not to confuse users. I think having a `core` export may be easier for a user to ignore than several functions, so that's why I've grouped the core utilities this way.

The test suite is actually the same one that I'd written in #236 and passes without any issue (yay)! In addition, I went and added information to the `docs` so folks can see instructions on the documentation site.

---

One thing to note is that I have **not** tested this change in a real application yet. My work here was purely porting the code from #236 in order to target `master` again. @cristianbote, might want to give 'er a try before we consider merging this.


benchmarks/perf-hash.js
```diff
@@ -1,31 +1,34 @@
 const Benchmark = require('benchmark');
```
jakehamilton, line 1: This file had a conflict, so when the pre-commit formatting run happened, it was included. No changes to this file that are important.

docs/docs/api/styled.md
```diff
@@ -6,15 +6,15 @@ sidebar_label: styled
```
jakehamilton, line 6: This file had a conflict, so when the pre-commit formatting run happened, it was included. No changes to this file that are important.

docs/docs/integrations/css-prop.md
```diff
@@ -27,7 +27,14 @@ List the plugin in `.babelrc`:
 Usage:
```
jakehamilton, line 27: This file had a conflict, so when the pre-commit formatting run happened, it was included. No changes to this file that are important.

jakehamilton:
It looks like the `filesize` check is failing, but I'm not sure why or how to fix that. Might need some help there 😅

jakehamilton:
This PR has sat around for a while so I'll be closing it. If anyone wants to pick this feature up, please feel free to use the code here!

PR #364, opened 2021-08-15
Fix forwardProps type

oliviertassinari:
The type doesn't seem correct, the docs: https://goober.js.org/api/setup. Noticed, while I was working on https://github.com/mui-org/material-ui/pull/27776.

### DIFF
--- a/goober.d.ts
+++ b/goober.d.ts
@@ -46,9 +46,7 @@ declare namespace goober {
         (props: any, ref: any): any;
     };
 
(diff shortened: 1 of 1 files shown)


cristianbote:
Oh @oliviertassinari that's a good catch! Thank you!

Also nice exploration! 😄  Love the chance you have given to goober. If there's any more feedback you would need to complete it or issues to resolve let me know and I'll make sure to get some time on it.

Have a great one!

PR #377, opened 2021-10-05
Fix type in readme

tenphi:
just a small typo


README.md
```diff
@@ -169,7 +169,7 @@ Coming soon!
 The benchmark is testing the following scenario:
 
 ```jsx
-import styled from 'package';
+import styled from 'goober';
```
cristianbote, line 172: Hya @tenphi! Thank you so much for opening this PR. 

This is totally my fault, I didn't knew how to explain this better, but I was trying to describe how I've tested each package. And that is what the `'package'` reference is pointing to. Do you think if this would be:

```js
import styled from 'packageName';
```
would make more sense in this context?

cristianbote:
```suggestion
import styled from '<packageName>';
```

PR #398, opened 2021-12-05
[feature] Remove empty values of properties.

ksenginew:
@cristianbote First, Thank you for this great library.

This feature can handle following case.
```js
const Btn = styled("button")((props) => ({
  borderRadius: props.rounded
}));

<Btn rounded="2px" />; // => { borderRadius: "2px" }

let isRounded = false
<Btn rounded={isRounded ? "4px" : null} />; // => { borderRadius: null }

<Btn>; // => { borderRadius: undefined }
```
This **should** work for something like this.
```js
const Btn = styled("button")(
  (props) => `
  border-radius: ${props.rounded};
`
);

<Btn rounded="2px" />; // => `border-radius: 2;`

let isRounded = false
<Btn rounded={isRounded ? "4px" : null} />; // => `border-radius: null;`

<Btn>; // => `border-radius: undefined;`
```

Thanks.

### DIFF
--- a/src/core/parse.js
+++ b/src/core/parse.js
@@ -25,7 +25,7 @@ export let parse = (obj, selector) => {
                 // Regular at rule block
                 blocks += key + '{' + parse(val, key[1] == 'k' ? '' : selector) + '}';
             }
-        } else if (typeof val == 'object') {
+        } else if ((typeof val)[0] == 'o') {
             // Call the parse for this block
             blocks += parse(
                 val,
@@ -43,7 +43,7 @@ export let parse = (obj, selector) => {
                       })
                     : key
             );
-        } else if (val != null) {
+        } else if (val != undefined) {
             // If this isn't an empty rule
             key = key.replace(/[A-Z]/g, '-$&').toLowerCase();
             // Push the line for this property
--- a/src/core/parse.js
+++ b/src/core/parse.js
@@ -43,7 +43,7 @@ export let parse = (obj, selector) => {
                       })
                     : key
             );
-        } else if (val != undefined && val != null) {
+        } else if (val != undefined) {
             // If this isn't an empty rule
             key = key.replace(/[A-Z]/g, '-$&').toLowerCase();
             // Push the line for this property
--- a/src/core/parse.js
+++ b/src/core/parse.js
@@ -12,52 +12,46 @@ export let parse = (obj, selector) => {
     for (let key in obj) {
         let val = obj[key];
 
-        // If this isn't an empty rule
-        if (val != undefined && val != null) {
-            // If this is a 'block'
-            if (typeof val == 'object') {
-                // If these are the `@` rule
-                if (key[0] == '@') {
-                    // Handling the `@font-face` where the
-                    // block doesn't need the brackets wrapped
-                    if (key[1] == 'f') {
-                        blocks += parse(val, key);
-                    } else {
-                        // Regular rule block
-                        blocks += key + '{' + parse(val, key[1] == 'k' ? '' : selector) + '}';
-                    }
-                } else {
-                    // Call the parse for this block
-                    blocks += parse(
-                        val,
-                        selector
-                            ? // Go over the selector and replace the matching multiple selectors if any
-                              selector.replace(/([^,])+/g, (sel) => {
-                                  // Return the current selector with the key matching multiple selectors if any
-                                  return key.replace(/(^:.*)|([^,])+/g, (k) => {
-                                      // If the current `k`(key) has a nested selector replace it
-                                      if (/&/.test(k)) return k.replace(/&/g, sel);
-
-                                      // If there's a current selector concat it
-                                      return sel ? sel + ' ' + k : k;
-                                  });
-                              })
-                            : key
-                    );
-                }
+        if (key[0] == '@') {
+            // If these are the `@` rule
+            if (key[1] == 'i') {
+                // Handling the `@import`
+                outer = key + ' ' + val + ';';
+            } else if (key[1] == 'f') {
+                // Handling the `@font-face` where the
+                // block doesn't need the brackets wrapped
+                blocks += parse(val, key);
             } else {
-                if (/^@i/.test(key)) {
-                    outer = key + ' ' + val + ';';
-                } else {
-                    key = key.replace(/[A-Z]/g, '-$&').toLowerCase();
-                    // Push the line for this property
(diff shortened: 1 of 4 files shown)


src/core/parse.js
```diff
@@ -8,52 +8,55 @@ export let parse = (obj, selector) => {
...
-                : key;
-
-            // If these are the `@` rule
-            if (key[0] == '@') {
-                // Handling the `@font-face` where the
-                // block doesn't need the brackets wrapped
-                if (key[1] == 'f') {
-                    blocks += parse(val, key);
+        // If this isn't an empty rule
+        if (val != undefined && val != null) {
```
cristianbote, line 16: I think we could just check for: `val != undefined` since `null` is also equal to `undefined` but not strictly equal. What do you think?
```suggestion
        if (val != undefined) {
```

src/core/parse.js
```diff
@@ -8,52 +8,55 @@ export let parse = (obj, selector) => {
...
-                outer = key + ' ' + val + ';';
-            } else {
-                key = key.replace(/[A-Z]/g, '-$&').toLowerCase();
-                // Push the line for this property
-                current += parse.p
-                    ? // We have a prefixer and we need to run this through that
-                      parse.p(key, val)
-                    : // Nope no prefixer just append it
-                      key + ':' + val + ';';
+                if (/^@i/.test(key)) {
```
cristianbote, line 49: Good refactor! You changed this to make it more compact or was there a size constraint? If I remember correctly I expanded the check into two to gain some more gzip size. But I might remember it wrong.

cristianbote:
@ksenginew this is an amazing find and change! 🙌 Thank you so much for this. Left two comments. Let me know your thoughts and this is good to go!

src/core/parse.js
ksenginew: @cristianbote can we use
```suggestion
        if (val != null) {
```

src/core/parse.js
ksenginew: I don't know? Can you test it or how can i test it?

ksenginew:
Thanks @cristianbote 

 I am developing an css framework(powered by atomic css-in-js library).

Goober is not designed for atomic design. So I am developing new one.

I learned (and copied) many thing from goober. When other libraries fighting with parsers like stylis(~10kb I think), You did it with two regular expressions.
 I'll include proper attribution for goober.

Thanks

src/core/parse.js
ksenginew: Now It's fixed

src/core/parse.js
```diff
@@ -8,53 +8,50 @@ export let parse = (obj, selector) => {
...
+                              // If the current `k`(key) has a nested selector replace it
+                              if (/&/.test(k)) return k.replace(/&/g, sel);
+
+                              // If there's a current selector concat it
+                              return sel ? sel + ' ' + k : k;
+                          });
+                      })
+                    : key
+            );
+        } else if (val != undefined) {
```
ksenginew, line 46: in https://github.com/cristianbote/goober/pull/398#discussion_r762567062

can we use
```suggestion
        } else if (val != null) {
```

src/core/parse.js
```diff
@@ -8,53 +8,50 @@ export let parse = (obj, selector) => {
...
-                          return sel ? sel + ' ' + k : k;
-                      });
-                  })
-                : key;
-
+        if (key[0] == '@') {
             // If these are the `@` rule
-            if (key[0] == '@') {
+            if (key[1] == 'i') {
+                // Handling the `@import`
```
ksenginew, line 18: https://github.com/cristianbote/goober/pull/398#discussion_r762567239

> The `@import` rule must be at the top of the document (but after any `@charset` declaration).
_from [w3schools](https://www.w3schools.com/cssref/pr_import_rule.asp)_

Using `@import` is highly discouraged.(HTML `link` is recommended)

Should we remove support for `@import`???(is it a breaking change?)

ksenginew (commented):
Rewritten

cristianbote:
That's great to hear @ksenginew! I think in the land of atomic css I believe https://github.com/kripod/otion is the one that I would highly recommend. Then of course [twind](https://twind.dev/) if you are more into tailwind like experience.

Anyways your changes actually shaved `-1B` 🙌 🥳  which is really really great! Thank you!

ksenginew:
Cool I had added a feature and shaved 1kb.
Can I join Grate shave off challange.

src/core/parse.js
ksenginew: According to [this MDN article](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness#loose_equality_using_) `null` is equal to `undefined`

ksenginew:
@cristianbote Now we can merge this pr?

cristianbote:
Hey @ksenginew, you did a great job in here! I need to check if there are any perf degradation with the parse changes. I'll get back.

Usually that means running the [styled-components benchmark](https://github.com/styled-components/styled-components/tree/main/packages/benchmarks) locally and the goober perf tests which you can find in the repo.

ksenginew:
Ah it's ok 😊

ksenginew:
@cristianbote What about the perf tests?

cristianbote:
@ksenginew it seems that there are no distinguishable regression here so we are good! 🙌 🥳 

Congrats on working on this and shaving off some bytes! 🎉 Wohoo! This is a great candidate for the [The Great Shave Off challenge](https://goober.rocks/the-great-shave-off) but I would like to reward your for your sustained effort so with this being said we want to reward you with $50 USD 🎉!
Go over https://opencollective.com/goober and submit an invoice for the above sum.

Again thank you so much for your involvement!

ksenginew:
Welcome @cristianbote (Also thanks for your reward.)

ksenginew:
I have an Idea, I can `styled.div` syntax to main goober package with one line of JavaScript . It works without babel plugin.
<details>
    <summary>Expand to see the solution</summary>

```js
// styled.js
// ...

/**
 * styled function
 * @param {string} tag
 * @param {function} forwardRef
 */
let styled = function (tag, forwardRef) {
    // ...
}

// this is the working part
styled = new Proxy(sty, { get: (target, tagName) => target(tagName) })

export { styled, setup };
```
Try the hacked [sandbox](https://codesandbox.io/s/preact-goober-forked-n91ln)

</details>

---

Can I open a PR?

PR #399, opened 2021-12-08
Move cache to styled function

### LINKED ISSUE #397: How to make every class name to be unique?
image

B-Teague:
Fixes #397
Still need to add a new integration test to make sure it generates new class names per styled component.


B-Teague:
Doesn't actually fix the bug.  The cache is initialized with the style function, so it uses the same cache for all styled calls.

PR #401, opened 2021-12-10
Replace hash function with internal counter

### LINKED ISSUE #397: How to make every class name to be unique?

B-Teague:
Saves about 16 bytes from goober by replacing the toHash file with a simple counter.
Faster performance because there is no more hashing.
Also reduces bundle sizes with short class names.
Also fixes #397 

Each call to style creates an internal "id"
The cache key is the compiled css with the id appended (defaults id to empty string for css, glob, and keyframes)

If the cache key is not found, then a new class name is generated using an internal counter.


cristianbote:
@B-Teague this is a neat idea! And I believe this is something similar with that I've seen implemented in `cxs` https://github.com/cxs-css/cxs. The main problem here is the hydrating part. The output based on counter is not gonna be the same on server and on the client. The whole idea of the hash based on the css block was to have a cache system and identification of previously used styles.
Basically what would happen in this case would be that the whole styles and classNames would generate again on the client. Not really what I envision for goober to be.

What do you think?

B-Teague:
Whoops, didn't even think about hydration which is super important with webpage caching and SEO.  Yes, we definitely want to always have the same css generated both service and client side.

Any ideas on how to resolve #397 then?

Profesor08:
@B-Teague 
> Any ideas on how to resolve #397 then?

One className for common cached styles. Another one unique for each component. And all relactions between components should use their unique classNames. Unique classNames will be enought to impement by incrementing starting from some static value, but with different prefix, to avoid collision with cached classNames. By this, server and client will make same output, just avoid random.

PR #402, opened 2021-12-10
Assign id to styled components

### LINKED ISSUE #397: How to make every class name to be unique?

B-Teague:
@cristianbote I believe this approach will work without causing a hydration issue.
This approach let's the user assign a unique id to their styled components which is used by the toHash function to generate different classNames per styled component.

```javascript
const Button = styled('button', null, 'id1')``;
const AnotherButton = styled('button', null, 'id2')``;
```

Fixes #397

### DIFF
--- a/src/__tests__/css.test.js
+++ b/src/__tests__/css.test.js
@@ -31,14 +31,28 @@ describe('css', () => {
 
         expect(compile).toBeCalledWith(['base', ''], [1], undefined);
         expect(getSheet).toBeCalled();
-        expect(hash).toBeCalledWith('compile()', 'getSheet()', undefined, undefined, undefined);
+        expect(hash).toBeCalledWith(
+            'compile()',
+            'getSheet()',
+            undefined,
+            undefined,
+            undefined,
+            undefined
+        );
         expect(out).toEqual('hash()');
     });
 
     it('args: object', () => {
         const out = css({ foo: 1 });
 
-        expect(hash).toBeCalledWith({ foo: 1 }, 'getSheet()', undefined, undefined, undefined);
+        expect(hash).toBeCalledWith(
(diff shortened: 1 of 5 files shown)


Profesor08:
> `const Button = styled('button', null, 'id1')``;`

This aproach is the same as `const Button = styled('button')`random_bytes_to_make_code_works_1278912391892`;`

There is an example of hot it works. Code blocks are editable.
https://emotion.sh/docs/styled

B-Teague:
@Profesor08 
> There is an example of hot it works. Code blocks are editable.
> https://emotion.sh/docs/styled

So looking at emotion, it does some fancy magic behind the scenes with styled components and uses the source code to generate unique class names using babel plugins.

I get the following error when referencing another styled component:
```javascript
import styled from '@emotion/styled';
```

Error
Component selectors can only be used in conjunction with @emotion/babel-plugin.

Using their provided macro, you see it uses the name of the components as part of the class name
```javascript
import styled from "@emotion/styled/macro";
```

```html
<div>
  <div class="css-pio4yf-Parent ew04m780">
    <div class="css-10c6c3i-Child ew04m782">Green because I am inside a Parent</div>
    <div class="css-1gn4z2i-ChildB ew04m781">Some Text</div>
  </div><div class="css-10c6c3i-Child ew04m782">Red because I am not inside a Parent</div>
  <div class="css-1gn4z2i-ChildB ew04m781">Some Text</div>
</div>
```

https://codesandbox.io/s/emotion-example-ekex6

B-Teague:
I'm closing this PR as there has been no more feedback.  I think we can state the library is working as designed unless you want to implement a way to generate unique class names for each time styled is called, but still preserve hydration for server side rendering.

Issue #207, opened 2020-12-07
2.1.0


cristianbote:
For `2.1.0` I wanna make room for other improvements and new features. Making room for them will require some breaking changes.

## Breaking changes
- Move `glob` into its own package: `goober/global-styles`;
- With the above, export a function(`createGlobalStyles`) that will return a `<GlobalStyles />` component that needs to be part of the rendered tree. Not that convinced about this one.
- Move `goober-autoprefixer` to: `goober/autoprefixer`;

## Improvements
- Add the `stylis` test suite.
`goober` it's already functioning on par with `stylis` and I would like to keep track of this with a separate test suite. See `stylis` branch.
- Fix the nasty `;` bug for base64 data #173 
The limitation here is the regex rule reads the `;` as end of css rule and we need to make sure to be greedy here and match the `;` that contains another one. Tricky.
- Handle the `ssr` async cases.
The `ssr` story works almost in all cases. I've ran and a few users reported that in some cases the shared `glob` styles for examples, or the in-memory cache it's not working as intended.
- Add `css` prop functionality
Since we already have the `pragma` function we could easily capture the `h` calls and overwrite it so we could catch the `css` prop and do the `styled` work basically.
- Improve the documentation
#187, #191 
- Create a landing page for this project

devrnt:
Great, `createGlobalStyles` sound very nice to me!

 I would like to see a solution for #203 as well, I've got a PR ready but this introduces a prefix-prop-naming (transient props), but as you already said you really don't want this prefix. Maybe you have something to work around this issue?

cristianbote:
@devrnt how's the size impact on your PR? 🤔  Of there's negligible size impact, we should adopt it honestly.

devrnt:
@cristianbote Size change is indeed negligible, just a few lines of additional code.

devrnt:
I've pushed a proposal https://github.com/cristianbote/goober/pull/210

PR #403, opened 2021-12-11
[FIX] base64 url in string is now working

### LINKED ISSUE #173: background image with base64 image url not working
Hi, background image with base64 data image is not working a expected
It can be easily reproduced by
```
import { css } from "goober";

const HeaderClass = css`
  color: white;
  padding: 1em;
  background-image: url("data:image/gif;base64,R0lGODlhEAAQAMQAAORHHOVSKudfOulrSOp3WOyDZu6QdvCchPGolfO0o/XBs/fNwfjZ0frl3/zy7////wAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAACH5BAkAABAALAAAAAAQABAAAAVVICSOZGlCQAosJ6mu7fiyZeKqNKToQGDsM8hBADgUXoGAiqhSvp5QAnQKGIgUhwFUYLCVDFCrKUE1lBavAViFIDlTImbKC5Gm2hB0SlBCBMQiB0UjIQA7");
`;

document.getElementById("app").innerHTML = `
<h1 class="${HeaderClass}">Hello Vanilla!</h1>
<div>
  We use Parcel to bundle this sandbox, you can find more info about Parcel
  <a href="https://parceljs.org" target="_blank" rel="noopener noreferrer">here</a>.
</div>
`;

```

ksenginew:
Fixes #173 
Also you can add a tick to #207 
You can write css without last semicolon. But goober wasn't supported. It's now working on goober.
eg:-
```diff
h1 {
  color: red;
-  opacity: 1;
+  opacity: 1
}
```
The regular expression is long but works finely. Inspired by 

https://github.com/reworkcss/css/blob/434aa1733f275a67ea700311451b98a14f8cc21a/lib/parse/index.js#L227

_Sadly_
**Size Change:** +105 B (2%) 

**Total Size:** 4.92 kB

| Filename | Size | Change | |
|:--- |:---:|:---:|:---:|
| `dist/goober.esm.js` | 1.21 kB | +25 B (2%) |  |
| `dist/goober.js` | 1.21 kB | +27 B (2%) |  |
| `dist/goober.modern.js` | 1.21 kB | +25 B (2%) |  |
| `dist/goober.umd.js` | 1.28 kB | +28 B (2%) |  |

<a href="https://github.com/preactjs/compressed-size-action"><sub>compressed-size-action</sub></a>


ksenginew:
@cristianbote 
I am sorry for this crazy regular expression. But it works. _maybe we can reduce the size._

**Size Change:** +129 B (2%) 

**Total Size:** 4.95 kB

| Filename | Size | Change | |
|:--- |:---:|:---:|:---:|
| `dist/goober.esm.js` | 1.22 kB | +31 B (2%) |  |
| `dist/goober.js` | 1.22 kB | +33 B (2%) |  |
| `dist/goober.modern.js` | 1.22 kB | +31 B (2%) |  |
| `dist/goober.umd.js` | 1.29 kB | +34 B (2%) |  |


<a href="https://github.com/preactjs/compressed-size-action"><sub>compressed-size-action</sub></a>

cristianbote:
Yeah the regex is complex now. 😬 I think I went through it a couple of times to get it and still haven't figured it out. Yet! I'll keep you posted. Another avenue with the regex is to handle the bg image differently.

ksenginew:
Hi @cristianbote I can explain the regex. Can we discuss to reduce the size of regex.

src/core/astish.js
```diff
@@ -1,4 +1,4 @@
-let newRule = /(?:([\u0080-\uFFFF\w-%@]+) *:? *([^{;]+?);|([^;}{]*?) *{)|(}\s*)/g;
+let newRule = /(?:([\u0080-\uFFFF\w-%@]+) *:? *((?:'(?:\\'|.)*?'|"(?:\\"|.)*?"|\([^\)]*?\)|[^{};])+)(?=[;}])|([^;}{]*?) *{)|(}\s*)/g;
```
cristianbote, line 1:
```suggestion
let newRule = /(?:([\u0080-\uFFFF\w-%@]+) *:? *((?:\([^\)]+?\)|[^{};])+)(?=[;}])|([^;}{]*?) *{)|(}\s*)/g;
```

Thanks for the ping! So deep diving into the rule was wondering why the need to target the `'|"` chars? 🤔 so did a quick test, remove them and tests are passing. So was there any particular caveat that you were chasing here?

ksenginew: Oh, I'll explain the RegEx. Also you can inspect this regex via https://regex101.com/r/MAfztK/1 link. It's pretty good.

cristianbote: With the above the changes are only `+16B` which is not bad at all.

cristianbote: @ksenginew yeah saw the regex101 explanation. Here's the regex101 example with the trimmed down rule https://regex101.com/r/3ylRk5/1

cristianbote: The initial regex rule matches the `'|"` and also the `(|)` braces. So that was my question if there's another case where this leads to issues if we remove the matching for `'|"`?

ksenginew: @cristianbote your regex fails for `content` property.
```css
.close-me::after {
    content: "closing with ; now it's closed"; /* failed here */
}
```

**My solution will not fail in rare cases. But your solution looks enough.**

This is our OLD regular expression.
```py
/(?:([\u0080-\uFFFF\w-%@]+) *:? *([^{;]+?);|([^;}{]*?) *{)|(}\s*)/g
```
We can split it -`.split("|")`
- declaration
    ```py
   (?:([\u0080-\uFFFF\w-%@]+) *:? *([^{;]+?);
    ```
- selector or at rule
    ```py
    ([^;}{]*?) *{)
    ```
- last `}` character
    ```py
    (}\s*)
    ```

We want to change the part of declaration
```py
(?:([\u0080-\uFFFF\w-%@]+) *:? *([^{;]+?);
```
- property
    ```
    (?:([\u0080-\uFFFF\w-%@]+)
    ```
- `:` mark. It's optional to handle `@charset`
    ```css
     *:? *
    ```
- value
    ```py
   ([^{;]+?);
   ```

We should choose value 
```py
 ([^{;]+?);
```

I had changed it to
```diff
- ([^{;]                                             +?   );
+ ((?:'(?:\\'|.)*?'|"(?:\\"|.)*?"|\([^\)]*?\)|[^{};])+)(?=[;}])
```
inside this group
```py
((?:'(?:\\'|.)*?'|"(?:\\"|.)*?"|\([^\)]*?\)|[^{};])+)
```
There are 3 uncaptured groups added.
```py
'(?:\\'|.)*?'
```
```py
"(?:\\"|.)*?"
```
```py
\([^\)]*?\
```
```py
[^{};])  # already existed
```
These 3 uncp... groups handles double quoted strings, single quoted strings and CSS function calls.
```py
'(?:\\'|.)*?'
```

Can we take a last decision. then I can optimize regex and commit it.

cristianbote: Oh, that makes sense now, those are for the `content` property which can have a variation of things inside it and not mention defined differently based on the input format.

Ok let's merge this and we can look for things to minify later on. Great work @ksenginew!

PR #408, opened 2021-12-30
fix: Correcting package type

### LINKED ISSUE #407: SyntaxError: Named export 'css' not found. (from esm import)
`import 'solid-styled-components'` causes the following error from goober:

```
import { css, setup as gooberSetup } from "goober";
         ^^^
SyntaxError: Named export 'css' not found. The requested module 'goober' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'goober';
const { css, setup: gooberSetup } = pkg;

    at ModuleJob._instantiate (node:internal/modules/esm/module_job:124:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:181:5)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:281:24)
    at async Module.run (file:///home/brian/work/ctx-core/ctx-core/node_modules/.pnpm/uvu@0.5.2/node_modules/uvu/run/index.mjs:10:3)
    at async /home/brian/work/ctx-core/ctx-core/node_modules/.pnpm/uvu@0.5.2/node_modules/uvu/bin.js:26:5
 ELIFECYCLE  Command failed with exit code 1.
 ELIFECYCLE  Test failed. See above for more details.
```

See https://github.com/solidjs/solid-styled-components/issues/10

rschristian:
Fixes #407 

If Goober is going to have the `"import"` condition set to use `.js` as the file extension (per #242, seems like there's some Webpack issues with `.mjs` still? Not sure what exactly you ran into), you'll need to set `"type": "module"` else Node is going to treat it like CJS every time.

I'm not sure why you were using `"browser"` for modern, and `"import"` for ESM? Everything that supports reading `"exports"` _should_ support modern output. If there's a reason for this I'm happy to revert it, though Microbundle will fight this a bit as it uses `"import"` to determine the modern output's name.

Had to bump Microbundle as it only recently got patched to support outputting `.cjs`.

Tested with Preact-CLI, WMR, Vite, CRA, and NextJS, everything seems to be working just fine.


cristianbote:
Hey @rschristian thank you so much for this! Alrighty looking over the changes my only worry is the new `.cjs` extension but I suppose everything is fine? Is there a node version where this could break though?

rschristian:
That's certainly a very real concern.

Everything should be fine in recent tools, and I tested a handful. There's all sorts of games we can play with this of course, correcting issues with older tools if they should be supported.

Of course, super happy to alter anything and work with you to test out build tooling versions to ensure compatibility. I only spotted the one issue related to this, and it's quite recent, so I figure Goober's usage in Node might be uncommon, so probably no rush is needed.

I'll convert this into a draft and try to go through some older versions of popular build tools after work tonight, see if I can find the breaking points for you.

> Is there a Node version where this could break though?

Not a currently supported version no. `.cjs` and `.mjs` have been supported and required since Node 12, which is the oldest LTS: https://nodejs.org/api/packages.html#type

I didn't spot the `"engines"` field or any other docs on what versions you want to support, so I took a bit of a guess and assumed 12 is okay. If not, the setup above could work I believe. Of course the `"engines"` field could also be added to enforce a minimum version/let users know theirs is outdated if you wish.

PR #412, opened 2022-01-11
fix: Microbundle builds

rschristian:
Sorry, my fault. I'm not sure if I was running Microbundle weirdly or if I just didn't see the output. As [you can see here](https://www.runpkg.com/?goober@2.1.3/global/dist/goober-global.cjs.js), subpackages were built with `.cjs.js` extensions due to a bug in older Microbundle versions.

Would really recommend moving to a mono-repo setup and install Microbundle just once. Managing 4 separate lock files at once seems a bit fragile. Microbundle has good support of being ran in other directories, as can be seen in [preact](https://github.com/preactjs/preact/blob/4f08ae4d607964ae8222e910d3b209def75626b0/package.json#L96). I haven't used NPM's workspaces, but I know Yarn (1)'s works quite well so I imagine NPM is a similar story.

### DIFF
--- a/global/package-lock.json
+++ b/global/package-lock.json
@@ -139,6 +139,247 @@
         "lodash": "^4.17.19"
       }
     },
+    "@babel/helper-define-polyfill-provider": {
+      "version": "0.3.0",
+      "resolved": "https://registry.npmjs.org/@babel/helper-define-polyfill-provider/-/helper-define-polyfill-provider-0.3.0.tgz",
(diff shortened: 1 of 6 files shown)


cristianbote:
@rschristian thanks for taking care of this one.

Yeah, each time I got into these kind of issues I think of "a better way to manage them". But then it passes and I don't have to touch them for a while. Thanks for your suggestions though.

rschristian:
Yeah totally get it. Easy enough to fix one-off issues like this anyhow.

PR #416, opened 2022-01-13
Revert "[FIX] base64 url in string is now working"

### LINKED ISSUE #415: Too much recursion error - v2.1.6
Using vite, solid-js, solid-styled-components, the following error occurs in the browser.

```
Uncaught (in promise) InternalError: too much recursion
t3 goober.modern.js:1
    c goober.modern.js:1
    u goober.modern.js:1
    get className index.js:38
    get dev.js:1258
    get dev.js:1229
    get dev.js:1295
    get dev.js:1229
    assign web.js:213
    spreadExpression web.js:342
    runComputation dev.js:697
```

```js
          let className = css.apply(
            { target: _ctx.target, o: append, p: withTheme, g: _ctx.g },
            args
          );
```

`css.apply` is being called.

https://github.com/solidjs/solid-styled-components/blob/main/src/index.js#L38-L41

cristianbote:
Reverts cristianbote/goober#403
Fixes #415


ksenginew:
I am very sorry @cristianbote . How can we fix this.

cristianbote:
Oh nothing to be sorry about @ksenginew I keep trying to zero-in on what exactly breaks in here https://codesandbox.io/s/romantic-pine-flz85?file=/src/index.js but couldn't figure it out and don't have much time now to look into.

cristianbote:
If only we could find what exactly breaks and add a test for it to not regress on that bad code.

ksenginew:
> If only we could find what exactly breaks and add a test for it to not regress on that bad code.

Yes, even I can't find the bug but it happens after(or around) following par of normalize.css
```css
button,
select { /* 1 */
  text-transform: none;
}
```

PR #418, opened 2022-01-13
Stringify object using Object.entries instead of separated function

RealPeha:
Shaved off quite a few bytes by removing the `stringify` function and using `Object.entries` instead. I don’t know how this affected performance, running the benchmark did not give me the opportunity to evaluate the difference.

Also now this stringify returns a slightly different result.
Example object:
```json
{
   "value": 10
}
```
Old version: `value10`
My version: `value,10`

This affected the hash, so I had to edit the tests

<h3>Size comparison</h3>
<b>Before:</b>

```
Wrote 1183 B: goober.cjs.gz
       1079 B: goober.cjs.br
Wrote 1188 B: goober.esm.js.gz
       1085 B: goober.esm.js.br
Wrote 1257 B: goober.umd.js.gz
       1137 B: goober.umd.js.br
Wrote 1188 B: goober.modern.js.gz
       1085 B: goober.modern.js.br
```

<b>After:</b>
```
Wrote 1167 B: goober.cjs.gz
       1062 B: goober.cjs.br
Wrote 1173 B: goober.esm.js.gz
       1067 B: goober.esm.js.br
Wrote 1239 B: goober.umd.js.gz
       1120 B: goober.umd.js.br
Wrote 1173 B: goober.modern.js.gz
       1067 B: goober.modern.js.br
```

### DIFF
--- a/src/__tests__/integrations.test.js
+++ b/src/__tests__/integrations.test.js
@@ -103,7 +103,7 @@ describe('integrations', () => {
                 '"',
                 ' ', // Empty white space that holds the textNode that the styles are appended
                 '@keyframes go384228713{0%{opacity:0;}99%{opacity:1;color:dodgerblue;}}',
-                '.go1127809067{opacity:0;background:cyan;}',
+                '.go2626112847{opacity:0;background:cyan;}',
                 '.go3865451590{color:red;}',
                 '.go3991234422{color:cyan;}',
                 '.go3991234422 .go3865451590{border:1px solid red;}',
@@ -111,9 +111,9 @@ describe('integrations', () => {
                 '.go3206651468{color:green;}',
                 '.go4276997079{color:orange;}',
                 '.go2069586824{opacity:0;animation:go384228713 500ms ease-in-out;}',
-                '.go631307347{foo:1;color:red;baz:0;}',
-                '.go3865943372{opacity:0;}',
-                '.go1162430001{opacity:0;baz:0;}',
+                '.go2648019963{foo:1;color:red;baz:0;}',
+                '.go3913223576{opacity:0;}',
+                '.go3863620777{opacity:0;baz:0;}',
                 '"'
             ].join('')
         );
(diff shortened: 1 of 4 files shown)


cristianbote:
Thanks @RealPeha! Will look into the perf to see if there are any regressions.

RealPeha:
Replaced `Object.entries(compiled) + ''` with `Object.entries(compiled) + 0` and shave 1 byte for umd and cjs bundles

index.d.ts
```diff
@@ -0,0 +1,3 @@
+export { extractCss } from './core/update';
+export { styled, setup } from './styled';
+export { css, glob, keyframes } from './css';
```
cristianbote, line 3: Oh this is generated by microbundle but I've added a flag to not get this anymore. Can you check that you have this line in your fork? https://github.com/cristianbote/goober/blob/master/package.json#L75

cristianbote:
@RealPeha thank you so much for your work on this! 🙏 I appreciate it!

Looked at the perf scores and this is the breakdown:

For node env:
```
> NODE_ENV=production node benchmarks/perf.cjs


Starting: styled:object
▸ goober (Object.entries) x 135,277 ops/sec ±3.09% (76 runs sampled)
▸ goober x 190,528 ops/sec ±2.55% (84 runs sampled)

Fastest is: goober

Starting: styled:tagged
▸ goober (Object.entries) x 206,359 ops/sec ±1.73% (92 runs sampled)
▸ goober x 210,974 ops/sec ±1.61% (90 runs sampled)

Fastest is: goober

Starting: styled:array
▸ goober (Object.entries) x 131,862 ops/sec ±2.71% (87 runs sampled)
▸ goober x 187,293 ops/sec ±2.30% (87 runs sampled)

Fastest is: goober
```
Changed locally the test so it'll head-to-head the changed version -- the one with Object.entries, with the regular master one. This is a pretty decent regression in terms of perf. This will affect SSR performance or build durations.

Now in terms of browser:
```
goober (Object.entries) Mount deep tree (500) = 11.28 ±01.11 ms
goober (Object.entries) Mount deep tree (500) = 11.76 ±00.51 ms
goober (Object.entries) Mount deep tree (500) = 11.66 ±00.61 ms

goober Mount deep tree (500) = 11.57 ±00.53 ms
goober Mount deep tree (500) = 11.53 ±00.52 ms
goober Mount deep tree (500) = 11.49 ±00.66 ms
```
There really isn't much of a difference.

What are your thoughts on this? On one hand goober does get smaller but at the cost of perf degradation on node env and on the other the browser based rendering is just fine.

index.d.ts
RealPeha: Yes, I have the same package.json, but this file is still being created

RealPeha:
I think that performance should be a priority, and reducing the bundle by a couple of bytes does not provide any benefit. In any case, I'm doing this for fun and I have a more ideas on how to shave off a couple of bytes, so there will be new PRs soon 😉

PR #421, opened 2022-01-16
Some tricks for shave bytes (-10B for esm)

RealPeha:
* update.js - Use bitwise XOR for indexOf check
* parse.js - Optimized check for !=undefined
* hash.js - Remove the extra condition and variable
* compile.js - Just moved the condition

### DIFF
--- a/src/core/compile.js
+++ b/src/core/compile.js
@@ -33,9 +33,9 @@ export let compile = (str, defs, data) => {
                 // Here we check if this is strictly a boolean with false value
                 // define it as `''` to be picked up as empty, otherwise use
                 // res value
-                tail = res === false ? '' : res;
+                tail = res;
             }
         }
-        return out + next + (tail == null ? '' : tail);
+        return out + next + (tail == null || tail === false ? '' : tail);
     }, '');
 };
--- a/src/core/hash.js
+++ b/src/core/hash.js
@@ -40,21 +40,28 @@ export let hash = (compiled, sheet, global, append, keyframes) => {
     let className =
         cache[stringifiedCompiled] || (cache[stringifiedCompiled] = toHash(stringifiedCompiled));
 
-    // If there's no entry for the current className
-    if (!cache[className]) {
-        // Build the _ast_-ish structure if needed
-        let ast = stringifiedCompiled !== compiled ? compiled : astish(compiled);
-
-        // Parse it
-        cache[className] = parse(
-            // For keyframes
-            keyframes ? { ['@keyframes ' + className]: ast } : ast,
-            global ? '' : '.' + className
-        );
-    }
-
     // add or update
-    update(cache[className], sheet, append);
+    update(
+        (cache[className] =
+            cache[className] ||
+            // If there's no entry for the current className
+            // Parse it
+            parse(
+                // For keyframes
+                keyframes
+                    ? // Build the _ast_-ish structure if needed
+                      {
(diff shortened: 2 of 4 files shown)


src/core/parse.js
```diff
@@ -43,7 +43,8 @@ export let parse = (obj, selector) => {
                       })
                     : key
             );
-        } else if (val != undefined) {
+        } else if (val + 1) {
```
ksenginew, line 46: `undefined + 1 = NaN` but `null + 1 = 1`. So you can use this trick to handle null values.

src/styled.js
```diff
@@ -48,15 +48,11 @@ function styled(tag, forwardRef) {
...
-            if (tag[0]) {
-                // Try to assign the _as with the given _as value if any
-                _as = _props.as || tag;
-                // And remove it
-                delete _props.as;
-            }
+            // Try to assign the _as with the given _as value if any
+            let _as = (tag[0] && _props.as) || tag;
+            // And remove it
+            tag[0] && delete _props.as;
```
cristianbote, line 55: Tried this one locally and increases the overall size 😵‍💫

src/core/update.js
```diff
@@ -17,5 +17,5 @@ export let extractCss = (target) => {
  * @param {Boolean} append
  */
 export let update = (css, sheet, append) => {
-    sheet.data.indexOf(css) == -1 && (sheet.data = append ? css + sheet.data : sheet.data + css);
+    ~sheet.data.indexOf(css) || (sheet.data = append ? css + sheet.data : sheet.data + css);
```
cristianbote, line 20: 🙌 this one shaves around 4B

src/core/parse.js
cristianbote: Nice!

src/core/hash.js
```diff
@@ -40,21 +40,28 @@ export let hash = (compiled, sheet, global, append, keyframes) => {
...
+                              stringifiedCompiled !== compiled ? compiled : astish(compiled)
+                      }
+                    : stringifiedCompiled !== compiled
+                    ? compiled
+                    : astish(compiled),
+                global ? '' : '.' + className
+            )),
+        sheet,
+        append
+    );
```
cristianbote, line 64: I believe this is where the shaves are huge. The large downside though is the readability and maintenance effort if we keep it in this form. I would rather see this as a last resort.

src/core/compile.js
```diff
@@ -33,9 +33,9 @@ export let compile = (str, defs, data) => {
                 // Here we check if this is strictly a boolean with false value
                 // define it as `''` to be picked up as empty, otherwise use
                 // res value
-                tail = res === false ? '' : res;
+                tail = res;
             }
         }
-        return out + next + (tail == null ? '' : tail);
+        return out + next + (tail == null || tail === false ? '' : tail);
```
cristianbote, line 39: For some reason this also increases the size.

cristianbote (commented):
Hey @RealPeha would you be open to only leave this two changes:
1. https://github.com/cristianbote/goober/pull/421/files#diff-990212c4bbe286410abd4e28bab4e1c8549abc5447855b21bcdabdd65f651dd2R46
2. https://github.com/cristianbote/goober/pull/421/files#diff-c172bd93077b152ff54176dc84cca295f93e5889871c70911ef02a4cbaaf3836R20

These two are shaving together close to 6B in my tests -- or more.

PR #424, opened 2022-01-19
skip parsing and formatting custom CSS properties

### LINKED ISSUE #423: Custom CSS properties are manipulated from camel to kabab casing
The camel-To-Kabab is just to explain what I'm seeing.

For the given global style:

```jsx
const GlobalVariable = createGlobalStyles`
  :root {
    --my-customProperty: red;
  }
`
```

The following will be applied to the browser:

```css
:root {
  --my-custom-property: red;
}
```

I don't think there are any rules being broken by my use of camel casing. The [MDN docs](https://developer.mozilla.org/en-US/docs/Web/CSS/--*#syntax) even highlight how casing is important in differentiating one property from another (see the "Note" in the link provided). But casing isn't mentioned as "an unallowed token".

Just from looking at the source and re-using what I think is the offender in my own app, I believe the issue to be occurring [here during parsing](https://github.com/cristianbote/goober/blob/92c6fa91f3dd60a77630276b5eff2731460048b1/src/core/parse.js#L48), but I'm unsure.

realStandal:
~*Untested*~

Closes #423 

Skips formatting custom CSS properties using a lax check that a property's key does not begin with `--`.

### DIFF
--- a/src/core/parse.js
+++ b/src/core/parse.js
@@ -43,7 +43,8 @@ export let parse = (obj, selector) => {
                       })
                     : key
             );
-        } else if (val != undefined) {
+            // Parse and standardize all but custom CSS properties
+        } else if (val != undefined && !key.startsWith('--')) {
             // If this isn't an empty rule
             key = key.replace(/[A-Z]/g, '-$&').toLowerCase();
             // Push the line for this property
--- a/src/core/parse.js
+++ b/src/core/parse.js
@@ -45,7 +45,7 @@ export let parse = (obj, selector) => {
             );
         } else if (val != undefined) {
             // Convert all but CSS variables
-            key = key.startsWith('--') ? key : key.replace(/[A-Z]/g, '-$&').toLowerCase();
+            key = /^--/.test(key) ? key : key.replace(/[A-Z]/g, '-$&').toLowerCase();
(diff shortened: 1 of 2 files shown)


realStandal:
(assuming CI passes) I've redone the checking to exclude replacing CSS variables so old tests pass, and have included some new ones.

I'm not sure of the consequences of parsing the variables, but it didn't cause me a problem with simple values 🤷‍♀️

src/core/parse.js
```diff
@@ -44,8 +44,8 @@ export let parse = (obj, selector) => {
                     : key
             );
         } else if (val != undefined) {
-            // If this isn't an empty rule
-            key = key.replace(/[A-Z]/g, '-$&').toLowerCase();
+            // Convert all but CSS variables
+            key = key.startsWith('--') ? key : key.replace(/[A-Z]/g, '-$&').toLowerCase();
```
ksenginew, line 48:
```suggestion
            key = /^--/.test(key) ? key : key.replace(/[A-Z]/g, '-$&').toLowerCase();
```

ksenginew (commented):
Shaved some bytes for you.

cristianbote:
Thank you @realStandal and @ksenginew!

PR #434, opened 2022-02-14
Update csstype peerDependency to v3.0.10

### LINKED ISSUE #433: Update csstype peer dependency package
Hello

We started to use react-hot-toast library. And you are their [dependency](https://github.com/timolins/react-hot-toast/blob/main/package.json).

And you have a peer dependency [csstype](https://github.com/frenic/csstype/releases) v^2.6.2, which is published at 29 Jan 2019

It seems like you used it only in your goober.d.ts, and I checked their updates, they added TTime into Properties interface generic for animations support and etc.

So the upgrade might be not heavy

Maybe you can update it to the latest version?

Thank you!

alexey-koren:
Resolve #433

### DIFF
--- a/package.json
+++ b/package.json
@@ -107,7 +107,7 @@
     "@types/react": "^16.9.34",
     "babel-jest": "^24.1.0",
     "benchmark": "^2.1.4",
-    "csstype": "^2.6.3",
+    "csstype": "^3.0.10",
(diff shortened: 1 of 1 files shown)


cristianbote:
This looks great @alexey-koren! Thank you so much for taking care of this. Merging.

PR #448, opened 2022-04-27
(fix) Replace newlines with empty space

### LINKED ISSUE #447: A newline will cause meaningful whitespace to be stripped.
This:
```
css`
background: #000
	no-repeat;
`
```
Gets outputted as this:
```
#000no-repeat
```

Same thing if it's spaces instead of tabs. If a newline is in the mix, the space/tab gets stripped also. Our project's code-formatting tool is inserting line-breaks creating invalid CSS when run through Goober v2.1.8.

cristianbote:
Fixes #447


cristianbote:
The size increase is kinda large but I'd like that the output is more standardised. So gonna merge this and release.

PR #453, opened 2022-05-12
Minified Files

awholegnuworld:
I minified most files in a way that would not affect performance.


awholegnuworld:
The large number of file deletions is because I couldn't for the life of me get Github to overwrite files via file upload.

cristianbote:
hey @awholegnuworld what's the reason behind this PR? 🤔 why minifying the source files would help reduce the size of the compiled library?

awholegnuworld:
Oh, I was unaware that it was compiled. This PR is useless then. 😅

Issue #451, opened 2022-04-27
v3


cristianbote:
Starting to think about `v3` but not necessarily new features that pops up but rather breaking changes in APIs.

1. Move away from second argument as function to object based to allow passing `forwardRef` and `shouldForwardProp` functions.
```jsx
// ❌ old
const Old = styled('div', React.forwardRef)`
  border-radius: 1em;
`

// ✅ new
const NewWay = styled('div', {
  forwardRef: React.forwardRef,
  shouldForwardProp: () => {}
})`
  border-radius: 1em;
`
```
2. With the above, the `setup` function will not be needed anymore to have a 4th argument
```js
// ❌ old
setup(React.createElement, undefined, undefined, shouldForwardPropFn)

// ✅ new
setup(React.createElement)
```
2. Remove `glob` export from `goober`

ksenginew:
HI @cristianbote I would like to introduce some of my ideas. (maybe it's a new implementation for goober)

currently goober converts input to an object(if input is an object keeps it as it is), then breaks down it to css strings, and update an style element(in DOM)

1. create a [constructable stylesheet](https://web.dev/constructable-stylesheets/). if it's not supported in the current environment, then import a very basic implementation(based on style element) and use it.
2. check if [css nesting](https://developer.chrome.com/articles/css-nesting/) is supported(its works for [60% of global users](https://caniuse.com/css-nesting)). otherwise import a parser to handle this behavior.
3. convert input to an css string.(without handling nested styles)
4. update stylesheet with styles.
5. walk through newly updated styles and prefix them.(it's now already parsed by browser)

any thoughts 👀

rafaell-lycan:
Hey @ksenginew the idea is not bad, but I'd say should be a opt-in/out feature that allow users to choose what flavor `extractCss` will output.

One thing it might be worth changing is replace `autoprefixer` with [Baseline](https://web-platform-dx.github.io/web-features/) which it's been widely adopted, specially on Vite.

There are room for improvements no doubt, and the direction @cristianbote and the community will take might make **goober** a no brainer choice for most projects migrating from Styled Components and Emotion.

PR #459, opened 2022-06-04
Feature/custom parser

ivoilic:
Summary:

This PR adds a new optional `parser` parameter to the setup function. If this function is passed in it overrides the existing parser function entirely.

Motivation:

I'm currently migrating a project from stitches to goober and wanted to customize goober to include some of the functionality that exists in stitches. Namely, the SCSS-like syntax stitches uses for referencing theme tokens. (Ex: `color: $primary;`) While it's totally feasible to fork goober to create a custom version, I feel this small change will open the door to others easily implementing custom versions of goober while still making use of the original package.

Changes:

- Converts `parse` into an object with the actual parser function as a child.
- Updates unit test to reflect the changes to `parse` (No new unit tests have been added).
- Adds a new optional `parser` parameter to the `setup` function which override the default parser function.
- Updates the docs to include basic info about the new `parser` parameter (No example right now though).

Breaking Changes:
- None 🎉

Note:

I'm probably missing something but I'm totally not seeing what is generating the types for this library, so they might need to be updated. Related to that, it would be helpful if the docs could be updated to provide more guidance on how to develop and contribute to goober!

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,10 +32,10 @@ It's a pun on the tagline.
 
 # Talks and Podcasts
 
-* [React Round Up](https://reactroundup.com/wrangle-your-css-in-js-for-peanuts-using-goober-ft-cristian-bote-rru-177) 👉 https://reactroundup.com/wrangle-your-css-in-js-for-peanuts-using-goober-ft-cristian-bote-rru-177
-* ReactDay Berlin 2019 👉  https://www.youtube.com/watch?v=k4-AVy3acqk
-* [PodRocket](https://podrocket.logrocket.com/) by [LogRocket](https://logrocket.com/) 👉 https://podrocket.logrocket.com/goober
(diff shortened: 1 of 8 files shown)


ivoilic:
Looking over my work and the docs again I've realized that some of the customization I wanted to achieve with this update can totally be done using a the function passed into the`prefix` parameter in `setup`. If you feel this change is too extreme I would suggest instead renaming the `prefix` parameter to something more general to indicate more clearly it can be used to manipulate the output for other reasons.

cristianbote:
woooow! This is amazing @ivoilic! Really like how smooth you've implemented.

I believe you are right, maybe `setup` needs another argument for the parser or as you said make the `prefixer` a generic parser function. Have you seen #451? Trying to collect a set of changes for a major v3 bump and I truly believe this could be a valuable add to the new version.

Let me know your thoughts and I think we can 100% include this.

PR #477, opened 2022-07-20
Fixed Styled Prop Warning

86Magical:
```js
// STYLED
const Btn = styled("button")`
  border-radius: ${(props) => (props.edge ? "9999px" : "4px")};
  background-color: #7920ff;
`;
```
 ```jsx
 //JSX
 <Btn edge>Hello CodeSandbox</Btn>
 ```

//Console Warning

Warning: Received `true` for a non-boolean attribute `edge`.

If you want to write it to the DOM, pass a string instead: edge="true" or edge={value.toString()}.
    at button

### DIFF
--- a/src/styled.js
+++ b/src/styled.js
@@ -1,15 +1,15 @@
-import { css } from './css';
-import { parse } from './core/parse';
-import {tag} from './core/domElement';
+import { css } from "./css";
+import { parse } from "./core/parse";
+import { tag } from "./core/domElement";
 let h, useTheme, fwdProp;
 function setup(pragma, prefix, theme, forwardProps) {
-    // This one needs to stay in here, so we won't have cyclic dependencies
-    parse.p = prefix;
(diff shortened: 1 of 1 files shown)


cristianbote:
Hey @86Magical, thanks for opening this PR. The fix is kinda lightweight `+6B` and I think it's a good fit. Can you look into the unit tests and change them as well? Afterwards let me know and we can merge and release this.

I do have some concern of compatibility issue, Is `props` property agnostic enough across React and Preact?

Thanks a lot!

86Magical:
YES ! Thank you so much for your email, I have gone through it and I
noticed my solution was wrong in a way.

If we use styled to style another component all props from the component
styled won't be available in the new styled-component

Eg

const ButtonTwo = styled(ButtonOne)`
 /*
in this case ButtonOne props are not accessible through ButtonTwo
*/
`

However I have been able to go through it and found a new solution for it.
I'll be uploading it ASAP.

Thank you Sir 💙

PR #479, opened 2022-07-21
Fixes For Styled Boolean Prop & Types for styled - tag param

86Magical:
I noticed that React & Preact would send a console warning when of if a boolean attribut is passed to any styled-component 

E.g 

```
const Btn = styled("button")`
  background-color: ${(props) => (props.primary ? "#7920FF" : "hotpink")};
  color: ${(props) => props.theme.color};
  border-radius: ${(props) => (props.rounded ? "9999px" : "5px")};
  padding: 5px 25px;
  width: 70%;
`;

```
JSX : ` <Btn primary> Hello Magical </Btn>`

Warning: ` Warning: Received `true` for a non-boolean attribute `primary`.

If you want to write it to the DOM, pass a string instead: primary="true" or primary={value.toString()}.
`

Also Created a new File domElements.js

in react i was able to achieve this by simply using React.ElementType to styled's tag param 

but since Goober isn't for react alone, i had to do it manually by typing each htmlElements

now autoComplete is possible while typing styled( ' button ' )


86Magical:
I really dont know what is causing ci/codesandbox to have issues building package

86Magical:
https://codesandbox.io/s/magical-cache-cyr2r8?file=/index.js

lib still works fine

PR #480, opened 2022-07-26
Feature/modular custom parser

ivoilic:
Summary:

This PR build on my [previous PR](https://github.com/cristianbote/goober/pull/459) by breaking the custom parser in modular functions.

Motivation: By splitting the custom parser into modular functions the end user can now replace only what they need to customize rather than needing to copy and paste large parts of goober's default parse function. IMO this strikes a nice balance between customizability and durability.

Changes:
- Replaces the prefix parameter in the setup function with a parser object parameter
- The prefix parameter becomes an optional value in the parser object
- The parse function is split off into 3 new functions that handle, @ rules, objects, and strings separately

Breaking Changes:
- The prefix parameter is no longer valid in the setup function. However, this could potential be mitigated by checking the type of the parameter and defaulting to the old prefixing behavior if the parameter is a function rather than an object.

### DIFF
--- a/README.md
+++ b/README.md
@@ -285,7 +285,7 @@ const Title = styled('h1', React.forwardRef)`
 `;
 ```
 
-### `setup(pragma: Function, prefixer?: Function, theme?: Function, forwardProps?: Function, parser?: Function)`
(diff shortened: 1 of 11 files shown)


ivoilic:
@cristianbote Any thoughts on this? I know you liked my work on the previous iteration of this concept.

PR #484, opened 2022-08-23
Fix #376

### LINKED ISSUE #376: Using createGlobalStyles together with themes leads to incorrect behaviour
When using a theme, if a button is used to change the background color for the body tag, for example, the first time it works fine, but the second time, it doesn't. Digging a little, I noticed that every time I click the button, changing the theme values, a new style tag is added to head. The first time it overwrites the default value, but once you want to go back to the default value, it was already overwritten and there is no change. I guess that the fix is updating the global style tag that was added by createGlobalStyles rather than adding a new style tag.

Minimal reproduction, replacing the code from the CodeSandbox for preact from [here](https://codesandbox.io/s/r15wj2qm7o).

```javascript
import "./style";
import { Component, render, h, createContext } from "preact";
import { useContext, useState } from "preact/hooks";
import { Result } from "./result";
import { styled, setup } from "goober";
import { createGlobalStyles } from "goober/global";

const Theme = createContext();
const useTheme = () => useContext(Theme);
setup(h, undefined, useTheme);

const SEARCH = "//api.github.com/search/repositories";

const Title = styled("h1")`
  text-align: center;
`;

const StyledResult = styled("div")`
  padding: 10px;
  margin: 10px;
  background: white;
  box-shadow: 0 1px 5px rgba(0, 0, 0, 0.5);
`;

const GlobalStyle = createGlobalStyles`
  html, body {
    background-color: ${({ theme }) => theme.a};
  }
`;

export default class App extends Component {
  constructor () {
    super();
    this.state = {a: '#ffff00'}
  }

  componentDidMount() {
    fetch(`${SEARCH}?q=preact`)
      .then((r) => r.json())
      .then((json) => {
        this.setState({
          results: (json && json.items) || []
        });
      });
  }

  render(props, { results = [] }) {
    return (
      <Theme.Provider value={{ a: this.state.a }}>
        <div>
          <GlobalStyle />
          <Title>Example</Title>
          <div class="list">
            {results.map((result) => (
              <StyledResult>
                <Result result={result} />
              </StyledResult>
            ))}
          </div>
          <button onClick={() => {
            if (this.state.a === '#ffff00') {
              this.setState({a: '#00ffff'});
            } else {
              this.setState({a: '#ffff00'});
            }
          }}>Change</button>
        </div>
      </Theme.Provider>
    );
  }
}

if (typeof window !== "undefined") {
  render(<App />, document.getElementById("root"));
}
```

Looky1173:
Resolves #376.

@cristianbote This PR would increase the bundle size of goober but only by a few bytes.
```
1242 B: goober.cjs.gz
1133 B: goober.cjs.br
1252 B: goober.modern.js.gz
1149 B: goober.modern.js.br
1252 B: goober.esm.js.gz
1149 B: goober.esm.js.br
1315 B: goober.umd.js.gz
1195 B: goober.umd.js.br
```

Here is a [CodeSandbox demo](https://codesandbox.io/s/theming-preact-goober-rk92xb) where `createGlobalStyles` works correctly with themes.

### DIFF
--- a/src/core/__tests__/hash.test.js
+++ b/src/core/__tests__/hash.test.js
@@ -40,7 +40,7 @@ describe('hash', () => {
         const res = hash('compiled', 'target');
 
         expect(toHash).toBeCalledWith('compiled');
-        expect(update).toBeCalledWith('parse()', 'target', undefined);
+        expect(update).toBeCalledWith('parse()', 'target', undefined, null);
         expect(astish).toBeCalledWith('compiled');
         expect(parse).toBeCalledWith('astish()', '.toHash()');
 
@@ -53,7 +53,7 @@ describe('hash', () => {
         expect(toHash).not.toBeCalled();
         expect(astish).not.toBeCalled();
         expect(parse).not.toBeCalled();
-        expect(update).toBeCalledWith('parse()', 'target', undefined);
+        expect(update).toBeCalledWith('parse()', 'target', undefined, null);
 
         expect(res).toEqual('toHash()');
     });
@@ -64,7 +64,18 @@ describe('hash', () => {
         expect(toHash).toBeCalledWith('global');
         expect(astish).not.toBeCalled();
         expect(parse).not.toBeCalled();
-        expect(update).toBeCalledWith('parse()', 'target', undefined);
+        expect(update).toBeCalledWith('parse()', 'target', undefined, null);
+
+        expect(res).toEqual('toHash()');
+    });
+
+    it('regression: global-style-replace', () => {
+        const res = hash('global', 'target', true);
+
+        expect(toHash).not.toBeCalled();
+        expect(astish).not.toBeCalled();
+        expect(parse).not.toBeCalled();
+        expect(update).toBeCalledWith('parse()', 'target', undefined, 'parse()');
 
         expect(res).toEqual('toHash()');
     });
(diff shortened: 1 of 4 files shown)


cristianbote:
Hey @Looky1173 thanks for opening this. Size-wise is fine. Can you add a test case for this?

Looky1173:
@cristianbote Oops, I forgot about the tests! 😅 I've just fixed the ones failing and added two additional tests.

Looky1173:
Hi @cristianbote, could you please approve the PR workflows so we could see if all tests pass? (They do pass on my local machine, they should do here too.)

cristianbote:
It's all looking great @Looky1173! Thanks for your contribution

gr0uch:
Hi, this causes a bug where if multiple instances of global styles are needed, the last one overwrites all, preventing multiple global styles from being applied.

I think this could be fixed by giving `cache.g` a unique key instead of `g`.

Looky1173:
Hi @gr0uch, how would we identify different invocations of `createGlobalStyles`? The only way I can think of is adding an optional second parameter to `createGlobalStyles` which could be used as a unique key.

Looky1173:
@gr0uch Or we could make use of [`_previousClassName`](https://github.com/cristianbote/goober/blob/master/src/styled.js#L31), passing it to the hash function.

gr0uch:
adding an optional second parameter would work. one other way would be to use `new Error().stack` to check if it is a repeated invocation of a previous function call, since the stack would be identical. 

the drawback is that `stack` is non-standard but in practice implemented by almost every browser.

Looky1173:
Since `createGlobalStyles` returns a component, we could also use the `id` prop of `<GlobalStyles />`. What do you think @gr0uch and @cristianbote?

gr0uch:
yeah allowing a prop to be passed would probably work, though less automatic than checking calling location via stack trace.

cristianbote:
I need to look into it to understand it better before having an opinion on it 🤔 give me a few

PR #492, opened 2022-10-08
Fix #491

### LINKED ISSUE #491: [bug] Falsy properties are not discarded
```jsx
const Title = styled("h1")`
  font-weight: bold;
  color: dodgerblue;
  background-color: ${(props) => (props.background ? "lightblue" : null)};
`;
```
When the `background` property is null or false, `background-color` is not stripped from the generated CSS.

image

[CodeSandbox demontrating the issue](https://codesandbox.io/s/goober-bug-falsy-properties-are-not-discarded-d3bh85)

Looky1173:
Fixes #491.

When `compile()` detects a falsy value, it replaces it with `_` which will be ignored later by `parse()`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "goober",
-  "version": "2.1.9",
+  "version": "2.1.10",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "goober",
-  "version": "2.1.10",
+  "version": "2.1.11",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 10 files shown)


Looky1173:
Hi @cristianbote, what do you think of this solution? It only adds a few bytes to the bundle size:
| Filename | Size | Change |
|:--- |:---:|:---:|
| `dist/goober.esm.js` | 1.26 kB | +11 B (0%) |
| `dist/goober.modern.js` | 1.26 kB | +11 B (0%) |
| `dist/goober.umd.js` | 1.32 kB | +10 B (0%) |

Looky1173:
Hi @cristianbote, any updates on this PR? 🙂

cristianbote:
Hey @Looky1173 I wanna take a deeper dive at this issue and figure out if we could avoid having to parse `_;` at all and rely on strict/shallow equality with falsy values.

Issue #545, opened 2023-05-11
`css()` taken by surprise when `this` is `window` (in “old-school” JS environment)


domq:
Fiddle at https://jsfiddle.net/2f6uzmb5/6/

Steps to reproduce:
1. Import Goober “the old-school way” e.g. with
   ```html
   <script src="https://cdn.jsdelivr.net/npm/goober@2.1.13/dist/goober.umd.js" />
   ```
2. Set `window.target` to something that internal function [`getSheet()`](https://github.com/cristianbote/goober/blob/master/src/css.js#L21) won't like
3. Call `css()`

Expected result: `css()` should return a fresh CSS class name.

Actual result:

```
TypeError: e.querySelector is not a function
```

(where `e` is presumably the minified variable that holds `window.target` inside `getSheet()`)

From reading the [relevant source code](https://github.com/cristianbote/goober/blob/master/src/css.js), it looks like setting any of `window.p`, `window.g`, `window.o`  or `window.k` could also result in funny behavior.

This doesn't happen in a modern (ESM) environment, as they [dropped the idea of squeezing the global object into `this`-less function calls](https://medium.com/codex/global-variables-and-javascript-modules-ce674a869164).

jluxenberg:
This will also happen when using `esbuild` to bundle, since it doesn't emit the "use strict" directive at the top of each bundled module.

See this esbuild-based repro: [2023-09-05_esbuild-x-goober_bug.tgz](https://github.com/cristianbote/goober/files/12533791/2023-09-05_esbuild-x-goober_bug.tgz)

and my fix: https://github.com/cristianbote/goober/pull/562

PR #562, opened 2023-09-06
fix: don't use `globalThis` when using `this`

jluxenberg:
Without "use strict", `this` will be set to `globalThis` which is typically the window object. If certain properties are set on the window object (e.g. "window.target"), this can cause problems for the `css` and `styled` functions.

Instead, check that `this` has been set to something other than `globalThis` before using it. This ensures that `this` will only have a value if it is explicitly bound, and otherwise will be `undefined`.

Note: I tried to add "use strict"; directives where needed, but couldn't get them to "survive" the build process -- they were removed. This fix seems to work just fine.

Closes https://github.com/cristianbote/goober/issues/545

### DIFF
--- a/src/css.js
+++ b/src/css.js
@@ -7,7 +7,7 @@ import { getSheet } from './core/get-sheet';
  * @param {String|Object|Function} val
  */
 function css(val) {
-    let ctx = this || {};
+    let ctx = (this !== globalThis ? this : undefined) || {};
     let _val = val.call ? val(ctx.p) : val;
 
     return hash(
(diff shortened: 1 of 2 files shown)


jluxenberg:
> Heya @jluxenberg,
> 
> Thank you so much for this PR and for the details. `goober` uses `microbundle` to bundle it up, `microbundle` has a `--strict` option. Have you tried passing that? Meaning here a `--strict` flag should be passed https://github.com/cristianbote/goober/blob/master/package.json#L76C72-L76C72

I did try that (and added in all the places where microbundle is invoked) but it didn't seem to change the build artifacts at all.

PR #575, opened 2024-01-10
export type in package.json should-forward-prop path

cindyvciandt:
Defines types in the exports map, so that imports work when leveraging the newer moduleResolution tsconfig values (nodenext, node16, bundler).

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,12 +44,14 @@
     "./global": {
       "import": "./global/dist/goober-global.modern.js",
       "require": "./global/dist/goober-global.cjs",
-      "umd": "./global/dist/goober-global.umd.js"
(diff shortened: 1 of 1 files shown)


cristianbote (approved):
Heya @cindyvciandt! Thank you for your contribution. Looks fine to me. Should we also add the `"types"` to `prefixer` and `global`?

cindyvciandt:
> Heya @cindyvciandt! Thank you for your contribution. Looks fine to me. Should we also add the `"types"` to `prefixer` and `"global"`?

Absolutely! I've already added the 'types' to the prefixer and 'global'. Check it out when you get a chance.

PR #589, opened 2024-08-29
Fix goober running in vitest

hpinkos:
Prevents an error when importing a component using goober into a vitest unit test
```
TypeError: Cannot add property p, object is not extensible
 ❯ a node_modules/goober/dist/goober.modern.js:1:2084
```

The error is happening because `this` is equivalent to `module.exports`
```
console.log(this);

[Module: null prototype] {
  css: [Function: u],
  extractCss: [Function: r],
  glob: [Function: bound u],
  keyframes: [Function: bound u],
  setup: [Function: m],
  styled: [Function: j]
}
```

Reproducible example of the error: https://github.com/hpinkos/vitest-goober

### DIFF
--- a/src/css.js
+++ b/src/css.js
@@ -7,7 +7,7 @@ import { getSheet } from './core/get-sheet';
  * @param {String|Object|Function} val
  */
 function css(val) {
-    let ctx = this || {};
+    let ctx = Object.isExtensible(this) ? this : {};;
     let _val = val.call ? val(ctx.p) : val;
 
     return hash(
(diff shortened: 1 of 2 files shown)


cristianbote:
heya @hpinkos thanks for looking into this. Adding that check adds `+17B` and  wondering if this https://github.com/cristianbote/goober/pull/562 could also fix what you are seeing but with a smaller footprint?

hpinkos:
@cristianbote unfortunately no. I did check that out and try it first, but in this environment `globalThis` is set up to be a mocked version of `window` so it's not equivalent to `this`

PR #595, opened 2024-10-06
Replace `JSX` with `React.JSX`

memark:
The global `JSX` type is deprecated in React 18.3 and removed in React 19 RC. This PR changes the code to use the supported `React.JSX` syntax instead.

### DIFF
--- a/goober.d.ts
+++ b/goober.d.ts
@@ -11,11 +11,13 @@ declare namespace goober {
 
     interface StyledFunction {
         // used when creating a styled component from a native HTML element
-        <T extends keyof JSX.IntrinsicElements, P extends Object = {}>(
+        <T extends keyof React.JSX.IntrinsicElements, P extends Object = {}>(
             tag: T,
             forwardRef?: ForwardRefFunction
         ): Tagged<
-            JSX.LibraryManagedAttributes<T, JSX.IntrinsicElements[T]> & P & Theme<DefaultTheme>
+            React.JSX.LibraryManagedAttributes<T, React.JSX.IntrinsicElements[T]> &
+                P &
+                Theme<DefaultTheme>
         >;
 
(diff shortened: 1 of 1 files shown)


cristianbote:
Heya @memark thanks for opening this PR. Changes look good, wondering though if this should be a major or minor update? 🤔 If it's minor then types will break for usual versioning?

memark:
This is the info I could find (which is a bit vague)
https://react.dev/blog/2024/04/25/react-19-upgrade-guide#the-jsx-namespace-in-typescript

Digging through the TS defs I find this

old
```ts
declare global {
    /**
     * @deprecated Use `React.JSX` instead of the global `JSX` namespace.
     */
    namespace JSX {
        interface IntrinsicElements {
            ...
        }
    }
}
```

new
```ts
declare namespace React {
    namespace JSX {
        interface IntrinsicElements extends GlobalJSXIntrinsicElements {}
    }
}

interface GlobalJSXIntrinsicElements extends JSX.IntrinsicElements {}
```

meaning that both references end up in the same place. So not a breaking change I would say.

cristianbote:
Thanks for looking into it @memark. Looks good, I'll merge and push a new version.

cristianbote (approved):

cristianbote:
Published `2.1.15`
https://github.com/cristianbote/goober/releases/tag/v2.1.15

JamiesWhiteShirt:
This change broke the documented workaround for getting types to work with Goober + Preact + Typescript. https://goober.js.org/features/typescript#note-when-using-preact

```ts
// preact.d.ts

import JSX = preact.JSX;
```

This variant seems to do the trick:

```ts
// preact.d.ts

namespace React {
  export import JSX = preact.JSX;
}
```

PR #597, opened 2024-10-11
(fix) Fixes the :has and other selector when used with multiple (fixes #596)

### LINKED ISSUE #596: :has selectors not correctly mangled
Given:

```css
label:has(input, select) {}
```

goober turns that into:

```css
.go3847911401 label:has(input, .go3847911401 select) {}
```

Expected:

```css
.go3847911401 label:has(input, select) {}
```

I suspect similar issues exist for `:is` and `:not`.

cristianbote:
Expand the parser to also include multiple selector inside of `foo:has(one, two) {}`

Fixes #596


cristianbote:
Adds `+8B` but I guess these are needed. 🤷

PR #612, opened 2025-10-02
Added Content Security Policy support

### LINKED ISSUE #471: CSP Nonce
We have a pretty strict [CSP-Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) for our site and we would like to use a library which consumes `goober`.  The problem is that goober will inject styles directly to the document and if you have a CSP with pretty much anything other than `'unsafe-inline'` the browser will reject the insertion with an error that looks like:

```
Refused to apply inline style because it violates the following Content Security Policy directive: "style-src 'self' 'nonce-xz5NmxIucN9byPuojlWd4'". Either the 'unsafe-inline' keyword, a hash ('sha256-47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU='), or a nonce ('nonce-...') is required to enable inline execution.
```

I don't pretend to know every that would need to be implemented to solve this issue (and certainly not the consequences for making them).

However, I can say with certainty that the issue is this chunk of code:
https://github.com/cristianbote/goober/blob/0f74be5079e01916158198d74c8e5872fa7b518c/src/core/get-sheet.js#L17-L20
```javascript
            Object.assign((target || document.head).appendChild(document.createElement('style')), {
                innerHTML: ' ',
                id: GOOBER_ID
            })
```

At the very least the `createElement` would need to be changed to:
```javascript
  document.createElement('style', { nonce })
```

The next question will be where does the nonce come from, which there could be several answers.

But before we get there, **I'd like to do a quick reality check... Would this be a series of changes `goober` is open / willing to make?**

We are at an impasse right now with the library we want to use (it must be pulled out).  So there are two options we can pursue:
1. Get a change implemented in `goober` and then work the library author to consume that change
2. Just go a different direction (implement a different library)

What are people's thoughts?

ndbroadbent:
Fixes #471

This PR adds Content Security Policy (CSP) nonce support to goober, making it compatible with strict CSP policies that require style-src 'nonce-...'.

## Changes

- Multi-source nonce detection: goober now automatically checks for nonces from three sources in priority order:
  a. Explicit setNonce() call
  b. Vite convention: <meta property="csp-nonce"> tag
  c. Webpack convention: window.__webpack_nonce__
- New export: setNonce() function for explicit nonce control
- Documentation: Added CSP section to README explaining all three methods

## API

```
// Method 1: Explicit control
import { setNonce } from 'goober';
setNonce('your-nonce-here');

// Method 2: Vite (automatic detection)
// <meta property="csp-nonce" nonce="your-nonce-here" />

// Method 3: Webpack (automatic detection)
// window.__webpack_nonce__ = 'your-nonce-here';
```

## Implementation

The nonce is applied to dynamically created <style> elements in get-sheet.js. When a nonce is detected, goober adds nonce="..." attribute to the style element, allowing it to pass CSP checks.

## Testing

Tested with strict CSP policies requiring style-src 'nonce-...' - no inline styles are blocked and all dynamically generated styles work correctly.

## Backward Compatibility

Fully backward compatible - existing code works without changes. Nonce support is opt-in via one of the three methods above

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,7 @@ It's a pun on the tagline.
     -   [Sharing Style](#sharing-style)
     -   [Autoprefixer](#autoprefixer)
     -   [TypeScript](#typescript)
+    -   [Content Security Policy (CSP)](#content-security-policy-csp)-
 -   [Browser Support](#browser-support)
 -   [Contributing](#contributing)
 
@@ -703,6 +704,18 @@ Usage:
 -   [x] `globalStyle` (via `glob`) so one would be able to create global styles
 -   [x] target/extract from elements other than `<head>`
 -   [x] [vendor prefixing](#autoprefixer)
+
+# Content Security Policy (CSP)
+
+goober supports Content Security Policy nonces for inline styles. Set `window.__nonce__` before loading the library:
+
+```js
+<script nonce="your-nonce-here">
+  window.__nonce__ = 'your-nonce-here';
+</script>
+```
+
+The nonce will be added to goober's `<style>` element.
 
 # Sharing style
 
--- a/package.json
+++ b/package.json
@@ -131,7 +131,7 @@
   "typings": "./goober.d.ts",
   "filesize": {
     "./dist/goober.modern.js": {
-      "gzip": "1300B"
+      "gzip": "1301B"
     },
     "./dist/goober.cjs": {
       "gzip": "1300B"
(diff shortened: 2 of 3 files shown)


ndbroadbent:
Hi @cristianbote, sure I will look into that! Sorry, I actually came here from https://github.com/timolins/react-hot-toast/issues/138 and https://github.com/timolins/react-hot-toast/pull/412 - I should have paid closer attention to this repo and what you were trying to do! I missed the fact that it has to be as small as possible.

We could do it with just the `setNonce` call in that case and leave it up to the caller to handle webpack/vite conventions. I'll work on that

ndbroadbent:
Hi @cristianbote, very sorry but I had messed up the implementation so it was still causing CSP errors! I've opened a new PR with the fix: #613

It's important to set the nonce attribute BEFORE appending the style to the head, otherwise you still get CSP warnings. I've tested it thoroughly now and it works!

PR #621, opened 2026-06-04
chore: reduce core bundle size

pseudolabel:
- reduce core dist gzip size by ~79 B per build on average, 236 B cumulative across esm/modern/umd
- keep global style replacement working after extractCss()
- verified Build Check and Compressed Size pass
-  mildly cursed


cristianbote:
nice work @pseudolabel! 😄 if the API and functionality stays the same and no perf degradation in benches, this is a great candidate and also you will qualify for the never ending _The Grate Shave Off_! https://goober.rocks/the-great-shave-off

PR #622, opened 2026-09-19
fix: replace 'export =' with named exports in .d.ts files for TS>=5.5…

### LINKED ISSUE #606: ECMAScript modules & TypeScript >=5.5 error
In a Vite project with `type="module"`, the `tsc` step of the build fails when using TypeScript >= v5.5. The issue is:

```
node_modules/goober/goober.d.ts:3:1 - error TS1203: Export assignment cannot be used when targeting ECMAScript modules. Consider using 'export default' or another module format instead.

3 export = goober;
```

Package versions:
- vite 6.1.1
- typescript 5.5.0
- goober 2.1.16

Not completely sure if this may be caused by another issue specific to my project though, as this has not been reported yet and TS 5.5 has been out for a while...

aidansheehan:
## summary

goober's .d.ts files use `export = X; export as namespace X;`, which TypeScript
rejects (TS1203) once a package has `"type": "module"` and the consumer uses
`moduleResolution: "nodenext"` or `"node16"`. this rewrites goober.d.ts,
global/global.d.ts, prefixer/autoprefixer.d.ts, and
should-forward-prop/should-forward-prop.d.ts to use named exports + `export as
namespace` instead - same fix as https://github.com/paralleldrive/cuid2/pull/116, which hit the same error

### fixes

- `export =` in .d.ts files breaks under `moduleResolution: nodenext`/`node16` when package.json has `"type": "module"` (TS1203)
- rewrites goober.d.ts, global/global.d.ts, prefixer/autoprefixer.d.ts, should-forward-prop/should-forward-prop.d.ts to named exports + `export as namespace`

### not fixed

- default import (`import goober from 'goober'`) still fails under nodenext — no `export default`, never worked, not in README, not touched here
- I think true recommended TS fix is separate .d.cjs/.d.mjs  for CJS/ESM but that's a bigger change and touches build/exports map so have stuck with this for now, let me know if we prefer that

## reproduction

this reproduces the error from #606 against published goober@2.1.19 + typescript 5.5:

```bash
mkdir repro && cd repro
cat > package.json <<'EOF'
{ "name": "repro", "private": true, "type": "module" }
EOF
npm install goober@2.1.19 typescript@5.5 csstype @types/react
cat > index.ts <<'EOF'
import { styled, setup } from 'goober';
const B = styled('button')`color: red;`;
EOF
cat > tsconfig.json <<'EOF'
{ "compilerOptions": { "module": "nodenext", "moduleResolution": "nodenext", "strict": true, "noEmit": true, "jsx": "react" }, "files": ["index.ts"] }
EOF
npx tsc
```

You will get error:
```
node_modules/goober/goober.d.ts:3:1 - error TS1203: Export assignment cannot
be used when targeting ECMAScript modules.
```

## test plan

- reproduced TS1203 against published goober@2.1.19 with typescript 5.5 + moduleResolution nodenext
- confirmed it's gone under both nodenext and bundler resolution
- `import { styled } from 'goober'` — pass
- `import * as goober from 'goober'` — pass
- `import goober from 'goober'` (commonjs + esModuleInterop) — pass, same as before
- `import goober from 'goober'` under nodenext — fails, expected: no literal `export default`, as far as I can see never worked, not documented, not touched by this PR
- `import goober = require('goober')` (commonjs/node resolution) — pass
- goober/global, goober/prefixer, goober/should-forward-prop subpath imports — pass
- npm run test-ts / npm run test-unit-core — pass, unchanged
- test-ts doesn't test fix — ts-tests/tsconfig.json has moduleResolution "node", repo's typescript devDependency (3.6.3) predates "nodenext" (haven't added but let me know if we want a test for this, i think it'd need to be testing packged code rather than src and likely with a bespoke webpack config)

You can use this script to walk through each one:
```bash
// // namespace import — pass
import * as goober from 'goober';
goober.styled('button')`color:red;`;

// // require-style — pass (module: commonjs, moduleResolution: node)
import goober = require('goober');
goober.styled('button')`color:red;`;

// // require-style — pass (module: commonjs, moduleResolution: node)
import goober = require('goober');
goober.styled('button')`color:red;`;

// addon subpaths — pass
import { createGlobalStyles, glob } from 'goober/global';
import { prefix } from 'goober/prefixer';
import { shouldForwardProp } from 'goober/should-forward-prop';
```

(NB: last one appears to need a proper tarball as i think the self-reference on goober/global breaks with a symlink)

fixes #606


aidansheehan:
@cristianbote not sure how to get the synck checks to run, let me know if I should do anything my side 🙂

cristianbote:
Heya @aidansheehan! Thanks for this! Looks good. Nice work!

Let me check what's up with the CI, but I'll approve and merge and cut a new minor release.

